PR #44, opened 2021-09-29
Create MultipleCIToTask.js

snamjosh:
Endpoint:

https://<<<YOUR_INSTANCE_NAME>>>.service-now.com/api/gmi/related_live_ci_to_task

Payload Details:

User posts a JSON object with the task number and an array of CIs to associate to that task. Methods to identify the CIs being inserted can be sys_id, serial, FQDN, name, and IP, with that order of preference. Typically, we will only expect 1 identifier per CI line. fields in the post body:

Task: Task number that we should associate the CIs to. Required field, fail and return error code if this is not populated.

query_string: optional field that allows the user to pass in an override query string that we will add while querying for CIs. could allow for teams to only find CIs owned by a certain team, active CIs, retired CIs, only CIs of certain classes etc. if this is not provided or is an empty string, by default we will only search for live systems.

cis: Array of CIs that should be associated to the provided task. In each CI, can provide one of several methods to identify the CI, including name, serial, sys_id, IP, or FQDN. This same logic can be extended to include more custom fields if required.

I am using u_cmdb_ci_status for verifying production CI. it can be OOB field ( install_status) of whatever the customer is using.

### DIFF
--- a/MultipleCIToTask.js
+++ b/MultipleCIToTask.js
@@ -1,0 +1,167 @@
+(function process( /*RESTAPIRequest*/ request, /*RESTAPIResponse*/ response) {
+
+    // implement resource here
+
+    var numInserted;
+    var reqBody = request.body.data;
+    var a = JSON.stringify(reqBody);
+    var parser = new JSONParser();
+    var parsed = parser.parse(a);
+    var str = '';
+    var str1 = '';
+    var j;
+    var qstring = reqBody.query_string;
+    if (qstring != '') {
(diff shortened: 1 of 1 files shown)


earlduque:
Please follow the folder structure by placing your code in a relative category and subfolder, along with a readme.md file to explain. See existing code-snippets for examples

earlduque:
Reopening so that you can add more commits to this fork-branch.

earlduque:
Closed as duplicate. See #45

PR #55, opened 2021-10-01
GlideAggregate Group Count Snippet + Client Script URL Parameter fixes

mak1A4:
Added new directory for GlideAggreate Snippets, which also contains a subdirectory for the getGroupCount snippet.


mak1A4:
Also added some fixes for "Client Scripts/Get Value from URL Parameter"

PR #79, opened 2021-10-01
Added client script to verify x things are checked before showing field

jacebenson:
I added a script used some, in cases where you need complex logic to to vet weather or not enough checkbox are checked.


earlduque:
great code snippet, thanks jace

PR #81, opened 2021-10-01
Serviceportalwidgets set catalog vars from url

jacebenson:
Added service portal widget folder with a widget that sets variables from the url for all items on sc_cat_item


earlduque:
Hey jace this pr has updates and files that aren’t part of your widget

PR #83, opened 2021-10-01
Validation using Regular Expression

azeezgaa:


earlduque:
Would you be able to update your readme so that the link is a relative link and not a hard coded link to your fork? There's also a misspelling.

PR #84, opened 2021-10-01
Service Portal Widget: Set catalog item vars by url

jacebenson:
This is a widget you add to the page with the catalog item form and it will attempt to set the variables by the url.


MGOPW:
This looks awesome! Thanks for the contribution :)

MGOPW (approved):

PR #93, opened 2021-10-03
Initial adding code-snippet CatalogUtils.js

mskoddow:
If something is not ok, I am happy about a corresponding feedback!

### DIFF
--- a/Script Includes/CatalogUtils/CatalogUtils.js
+++ b/Script Includes/CatalogUtils/CatalogUtils.js
@@ -1,0 +1,87 @@
+var CatalogUtils = Class.create();
+
+/**
+ * Helper class for dealing with catalogs and catalog items.
+ * All methods can be called both server-side and client-side.
+ *
+ * @class CatalogUtils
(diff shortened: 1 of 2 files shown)


MGOPW:
Hi there! @mskoddow, Would you mind adding in more info about what your script does in the top section of your ReadMe.md file please? It isn't readily apparent what your script's purpose is and just saying "handling everything with service catalog" doesn't quite explain it's use, where to use it, why to use it, etc. 

Thanks :)

mskoddow:
Hi @MGOPW 

Thanks for the feedback! And you are right, the description is not very helpful at the moment, and it will become more clear in the next commits. The background is that I want to bundle my code snippets which are spread over several customer projects into a single helper class. Furthermore, I need this class as prerequisite for additional functionalities based on it (cloning, templating) in the service catalog area.

To be honest, I have only ever used Git in push mode and have never worked in a pull request. So hopefully it's okay modifying the readme.md file in this pull request.

Kind regards
Maik

PR #115, opened 2021-10-04
Export Kb Articles to PDF

snamjosh:


earlduque:
While we appreciate the contribution. This repository is for code snippets. Not update set dumps. Please reformat your commits to show actual code and make sure your readme has a file type like .md
Folder names should be based on the code type used.

snamjosh:
How will i commit objects in repo objects with no code like catalog items, record producers, flow designer flows?

earlduque:
One. You don’t even explain what it is or what the objects are because again you didn’t include a readme file as per outlined by the contributing guidelines. 
Two. If it’s not an actual code snippet then it doesn’t belong in this code-snippets repository.

PR #118, opened 2021-10-04
Fix Script: Install Base PDI Plugins

iar42:
Fix Script to speed up installation of multiple plugins on a fresh PDI (even with demo data).


earlduque:
This is awesome!

PR #120, opened 2021-10-04
minutesAgoStart & minutesAgoEnd

ReginaldFawcett:
Also a change to userID - change the example sys_id


earlduque:
Thank you @ReginaldFawcett for your contributions, would you edit the script files and add .js to the end of them so that GitHub's formatting takes effect when viewing the files?

earlduque:
fyi, you can add commits to open pull requests! I didn't close it so you could continue editing.

PR #125, opened 2021-10-04
AbstractStrategy

mr-swann:
in order to centralise the strategy and provide a common baseline, introduced AbstractStrategy and two ExampleStrategy APIs to demonstrate better how to implement this pattern

also updated the call to strategy.run() to return


earlduque:
Hey @mr-swann cool contribution but just remember the spirit of "code-snippets". We're starting to stray far from the idea of "Snippet is a programming term for a small region of re-usable source code, machine code, or text"

PR #131, opened 2021-10-05
Added a fun batman animation for portal

PranavBhagat18:


PranavBhagat18:
HI @earlduque , I updated it as per your feedback :)

PR #139, opened 2021-10-05
fixed some syntax errors + minor refactoring

mak1A4:


earlduque:
Hey @mak1A4 I very much appreciate the re-formatting/beautifying of a lot of the snippets, but we want to avoid complete refactoring of people's code in case there are use-cases for how it could be used that way. I'd much rather you submit a "variant" code-snippet for each of these re-factorings, a file that exists alongside the original file in the same folder (like scriptalternate.js for example).

mak1A4:
@earlduque okay thats fair, fixing the formatting and syntax errors is okay?
Also I was thinking about restructuring the repo, in my opinion it looks a little bit messy currently.
```
- Application Files
  - Business Rules
  - Script Includes
  - UI Actions
  - Scripted Rest APIs
  - Scheduled Jobs
  - Service Portal Widgets
  - ...
- API
  - GlideRecord
  - GlideAjax
  - GlideAggregate
  - GlideSystem
  - ...
- Snippets
  - Client
  - Server
```
imho something like this would give it a little bit more structure, what do you think?

PR #143, opened 2021-10-05
Floater widget & Currency validation catalog client script

ishaanshoor:


kcimpulse:
Accepting Pull they are in correct folders and structure,  in the future please expand your readme entries to elaborate a bit more, but I like the use of the screenshot, you might want to call out what in the screenshot is actually being referenced like a circle around i think your are using it for "feedback"

PR #145, opened 2021-10-05
Add Attachment icon in list view

kmohit2809:
If there is any attachment present on any record then add an attachment icon in list view of the record.


smartcarleen (changes requested):
Hi Mohit! Just one spelling correction! I think you meant to say "Voila" instead of "Volla". I believe you can make changes to the current pull request, you should not have to delete and resubmit!

kmohit2809:
Hi Carleen, 
I have completed the requested change. Hope it looks good now.

smartcarleen (approved):

PR #146, opened 2021-10-05
UI Action: Create Update Set on DEV

iar42:
A client UI Action for the Story form that opens up a new browser window with the Create New Update Set form on a specified DEV instance with Update Set name pre-filled with the Story number and short description.

Helps reducing copy/paste work and to keep up with Update Set naming standards.


smartcarleen (approved):
Great job @iar42!

PR #147, opened 2021-10-05
Inbound Actions - Advanced Scripts

ishaanshoor:
These inbound action scripts will help you with the following:
1) Creation - Map values from an email to variables and create a requested item for the same.
2) Updating - Check with already created requested item and update the variables and restart the workflow.
3) Cancellation - Check with already created requested item and cancel the same also cancel the workflow.


niamccash:
@ishaanshoor 

- The variable on line 4 of `cancel_catalog_item_from_email.js` doesn't seem to be in use in that script. Is it still required?
- Can you provide some more info in the readme and/or in the code about how other users might adapt these scripts for their specific variables?

ishaanshoor:
@niamccash 

Thanks, Nia for reviewing my pull request. 

I've removed the variable on line 4 of cancel_catalog_item_from_email.js 
Added more details in the readme.md file. Please let me know if anything else is required.

Cheers,
Ishaan

PR #150, opened 2021-10-05
Save Email Text as Attachment

azeezgaa:


earlduque:
Hi @azeezgaa it appears that you have two code snippets in this pull request. Just like update sets, we don't approve pull requests if the description of the request has file updates that appear to not belong in the request. You can either update the title to reflect the two code snippets included, or add additional commits to this pull request to reflect only the indicated snippet. Once this is done I can approve it.

azeezgaa:
Hi @earlduque 

I just mention the sample way of calling a script include ? so we cannot have that in readme file ?

PR #156, opened 2021-10-06
Base64 file encoder

ovarn:
Convert attachment to Base64


smartcarleen (approved):
Hi @ovarn! Thanks for this submission! We are going to approve, but I would encourage you to enhance your readme later on with information about when someone might use this script and why.

PR #164, opened 2021-10-06
Add custom greetings in portal home page

kmohit2809:
Add customize greetings in portal homepage bases on user's time zone.


kmohit2809:
Hi Carleen, I have updated the files and made the folder name more descriptive. Please let me know if there is anything else required from my end. Thanks

PR #165, opened 2021-10-06
Change Notification Preferences by Category

23Thom:


earlduque:
Hey @23Thom good submission, do you wanna break up your script file into multiple files?
for example:
notification_preference.js
can be:
serverscript.js
clientscript.js
style.css
code.html

Breaking it up into designate files with the correct file type allows github to display the script to the user with proper formatting according to the kind of code.

PR #196, opened 2021-10-08
Created API examples for the MIDServer class

oasif:
There is no documentation about the DiscoveryRange type, therefore I could not give examples to getForPing(DiscoverySchedule schedule, DiscoveryRange range)


earlduque:
@oasif Please note that the file structure still needs to be updated.
Your files are currently: 
MIDServer/readme.md

it should be something like:
MIDServer/API Class Examples/readme.md

oasif:
Thanks @earlduque, It was not clear how to structure the directories so I followed your suggestion

PR #208, opened 2021-10-09
Create a copy of incident in another instance

sumanthdosapati:


niamccash:
Is it possible to make the link in the readme file relative so it doesn't point to your personal github repository? 

I'm being picky here but please consider not shortening ServiceNow to 'SNOW', as there is another software out there called snow.

PR #224, opened 2021-10-10
Generic Script include function to get count of records matching query and calling script include as example.

ashu759:


niamccash:
Hey @ashu759, I'm having some troubles with the GenericNOW script include. Can you help me understand it, please?

Should there be a `:` after getcount function instead of `=`? My PDI complains about this when I paste the script in. 
And I think the script is missing closing brackets at the end of it. 

Also, I don't understand what line 1 of `var GenericNOW, getcount` is doing. It's not terminated with a `;` ?

ashu759:
Hey Nia, i see i did a typo. I have now this tested. So i will edit this and submit it.

niamccash:
Works now. Thanks @ashu759

PR #230, opened 2021-10-11
Renamed Mark Records Inactive - List Action

kmohit2809:
Removed the empty List action folder and renamed the file.


kmohit2809:
Hi @earlduque  and @smartcarleen this was an issue. This folder was following two tire structure because of which one folder was empty. So I have renamed the file and empty folder got removed. Please review it accordingly

PR #231, opened 2021-10-11
Format Phone number

PranavBhagat18:


PranavBhagat18:
@earlduque let me know if I need to make any other changes

PR #232, opened 2021-10-11
Adding a Scheduled Script Execution to lock out inactive users

R3J3NT:
Script to **lock out** users who have not logged into the system **longer than 30 days** and corresponding _readme.md_ file with description _(including two images - which are screenshots used in readme)_.


niamccash:
Your images are linked to files that are in your personal repository. Could you update the links to be relative so that they don't point to your personal repo, please? 

You can see an existing example of how it's done in this directory: https://github.com/ServiceNowDevProgram/code-snippets/tree/main/UI%20Macros/Purchase%20Order%20Approval%20Summarizer

PR #263, opened 2021-10-14
Added sample ATF test step using Jasmine for testing Script Includes

RahmanM:


niamccash:
Hi @RahmanM. Thank you for updating and adding the missing code. Can you please update the folder structure so that it's 2 levels instead of 3 eg. `Level_1/Level_2/files_here`

PR #264, opened 2021-10-14
Attachment to base64 in scope

23Thom:
Added a script to convert a attachment to base64 in a scoped application using the GlideSysAttachment API.


niamccash:
I'm not seeing the readme.md file that goes along with the attachment code in this PR. Can you please add a readme file?

Would be a bonus to see something in the readme about how this differs from the existing https://github.com/ServiceNowDevProgram/code-snippets/tree/main/Attachments/Attachment%20to%20Base64

PR #265, opened 2021-10-14
added 2 ways of nesting where/orWhere queries on the incident table

thorn91:
This example shows how to nest where and orWhere queries in a GlideQuery in two different ways.  It is testable in any default Personal Developer Instance portal.  These serve as a way of creating otherwise ambiguous queries without using an encoded query, which are machine readable but often not easy for humans to decipher which AND and ORs go where.


thorn91:
Please ignore - messed up my own commit history and am fixing.

PR #274, opened 2021-10-15
File Upload widget

rafzk:
This widget is using the 'Attachment API - POST /now/attachment/file' to upload multiple files on form submit.

The result will appear in the console or check directly the record.


rafzk:
image

PR #297, opened 2021-10-17
Create Choice Sets for unassociated choice table:element pairs

dean-croaker:
Business Rule

When: Before
Insert: True

When adding a new choice, and if there is currently no 'choice set' associated with table and element in the sys_choice table, the entries are marked as a DELETE operation in the update set, and all existing choices are also deleted.

This script ensures that a 'Choice Set' will be created for new Choices that don't currently have an associated Choice Set.


dean-croaker:
I had already submitted a pull request for the BR part of this, and then when I went to submit a second pull request to cover the Client Script - something went wrong. I don't know how to fix it and they are all bundled into the same pull request with some additional code modifications that were not (I thought) anything I had done.
This is my first pull request, so if there's something I can do to fix this and submit both separately, that would be great. Everything I tried seemed to make things worse or make no difference, so I stopped at risk of breaking anything further.

PR #310, opened 2021-10-18
Recursive Business Rule

aatrey882:
Recursive BR helps to automatically add the all affected CIs when Configuration item gets selected on Incident.
2. Glide Record Util(helps to get fields using record Util) and JSUtil validations


aatrey882:
This pull request holds Recursive Business rule, which helps to add all affected CIs when cmdb_ci gets selected, 2. Glide Record Util and JSUtil validations

PR #389, opened 2021-10-27
Is Time Up Function to keep execution time in check.

DynamiteC:


earlduque:
Not ServiceNow related

PR #394, opened 2021-10-28
Send event

anguspalmer:
Script to test event management configuration by sending test events using the Table API.


earlduque:
Not relevant to this _ServiceNow_ repository, closing.

PR #398, opened 2021-10-28
Adding a Background Script, to quickly generate events statistics

R3J3NT:
Background script, which **generate statistics about today's created events based on provided parameters** and corresponding _readme.md_ file with description _(including two images - which are screenshots used in readme and linked using relative path)_.


niamccash:
I tried running the script but the events average processing duration grouped by table name shows me all null values like:

`[Events stats] ------ Event in table: sc_catalog: null avg processing_duration`

What am I missing here?

R3J3NT:
You were right, @niamccash! I checked it and turns out it was my mistake; I pushed an older version which was not working correctly for examples with different fields for group by than aggregate function. It is fixed now in 9175fb6. An example correct result after execution on PDI:

`[Events stats] ------ Event in table: pa_job_logs: 241.00000000 avg processing_duration`

Thanks for that catch!

niamccash:
Runs fine now. Thank you for fixing @R3J3NT

PR #440, opened 2021-11-04
Dynamically create reference records when utilizing the AzureAD user provisioning integration with ServiceNow

TheTreyKyle:
Provides an example Business rule and explanation behind the following HI article; 
https://support.servicenow.com/kb?id=kb_article_view&sysparm_article=KB0779148

### DIFF
--- a/Attachments/ExportRecordsToCustomJSON/readme.md
+++ b/Attachments/ExportRecordsToCustomJSON/readme.md
@@ -1,0 +1,26 @@
+********************************************************************************
+Script include "GenerateJSONHelper"
+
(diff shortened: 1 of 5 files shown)


MGOPW:
Hello! Could you please review the following and make any modifications necessary? 

- Remove the XML from your files and just provide the scripts.
- Add the HI Article URL provided in your comment as part of the ReadMe 
- Provide a blurb at the top of what your scripts are used for/useful for and why i'd want to use them. Just in case I somehow miss the HI article link or am confused about what the context is. 

Thank you!

PR #446, opened 2022-09-30
New browser category and quick notes bookmarklet

johndahl-now:


johndahl-now:
I did not recognize any category that matches a browser code snippet. The only example listed in Utils appears to be for Node. Where should browser bookmarklets go?

Thanks.

PR #458, opened 2022-10-01
Compare roles

ttoter:
Compares the assigned roles between two users. The script will print roles that are shared and roles that are not shared between the two users.


earlduque:
useful!

PR #467, opened 2022-10-02
This script helps to get Table Name and related field Names in ServiceNow.

ynr-ram:
After executing this script the users can get all details of table names along with field names in servicenow.. when stakeholders want to know suddenly (or) at the time of integrations this script will help a lot and also we can modify the code to restrict it to a single table.


earlduque:
Reverted, again you need to follow the folder structure.

PR #470, opened 2022-10-02
Lock file

ynr-ram:
This script contains two parts..
1st part is client script and 2nd part is script include.. and this helps to lock the file till the developer completes his development, so that we can avoid concurrency issues.
Ex: If a "Developer A" working on "Widget A", parallelly if "Developer B" working on "Widget B" and each developer don't know that other developer is working on it, then if Developer B moves the code to higher environment before developer A completes his development.. then in such scenarios it creates the issue.. to avoid these kind of concurrency issues.. we can use locking system for each record.. in the above script I shown example for widget records.


ynr-ram:
Consider "LockYourFile.js" and "readme.md" for this contribution.. rest other two files are already contributed and merged to main one.

earlduque:
Please see comments on your last pull request. It was reverted because you didn't follow the folder structure, you must follow the requirements mentioned in the contributing.md file. 

You again didn't follow the structure here. 

Closing this without merge.

PR #476, opened 2022-10-02
Glide date get next Monday date

Danmax:
Here is a function using GlideDateTime to get the next Monday's Date.


Danmax:
Updated read me and js example log.

PR #497, opened 2022-10-06
add dark-mode-switcher ☀️/🌑

hrobarikt:
This is a simple demonstration of achieving a dark mode switcher for Employee Center portal. The switcher itself allows user to quickly toggle between multiple color modes.


earlduque:
Cool!

PR #507, opened 2022-10-11
Enhanced standard ticket page action widget

loudigi:
An enhanced standard ticket page action widget that adds approval functionality from within the Service Portal page along with the ability to navigate to the parent Request if the current record is a Request Item.


earlduque:
Usually we do not prefer XML payloads as a file in this repo as it makes it more difficult for someone to reference it as a code snippet, but an exception is being made here do to the better documentation of this one.

PR #511, opened 2022-10-13
Dynamic subcategory

allam1234:


allam1234:
I'm submitting a script include and a client script. This use case is for loading the subcategory options based on the category chosen. If there are no options available in the subcategory for the chosen category then the subcategory field will be hidden. I have created a onChange client for calling the script include by passing the category value so that the script include returns array of subcategory values for the given category.

earlduque (changes requested):
Hi @allam1234, thank you for contributing, please follow the folder+subfolder guidelines in the contributing.md file. Once you add a subfolder, I can approve this PR.

earlduque (changes requested):
Hi @allam1234, the subfolder should be describing the functionality of your code-snippet. Take a look at the existing folders and subfolders in this repository to see what I mean.

PR #526, opened 2022-10-14
Generate jwt token

urspvs:
This code snippet is for generating the JWT token from ServiceNow AES using Background Scripts module


earlduque:

PR #527, opened 2022-10-14
Convert date time

urspvs:
This code snippet is for convert the data time from one timezone to other from ServiceNow AES using Background Scripts module


earlduque:

PR #529, opened 2022-10-14
Get instance db size

urspvs:
This code snippet is for getting the DB size of an instance from ServiceNow AES using Background Scripts module


earlduque:

PR #551, opened 2022-10-20
Adding EfficientGlideRecord, BenchmarkRunner, TimeZoneUtils, and GetJournalEntryAsHTMLWithoutHeader

thisnameissoclever:
Adding EfficientGlideRecord to GlideAjax directory.


earlduque (changes requested):
Hey @thisnameissoclever, technically speaking I can't approve this Pull Request because the files you have added, despite being valid and good contributions, are mostly out of scope to what your Pull Request description says you are adding.

You can either:
- update your PR description to expand the scope
- remove the out-of-scope additions and submit them in their own Pull Requests.

thisnameissoclever:
@earlduque I confess my ignorance as to the intricacies of doing something like this via Github directly, but what I did was to submit a second pull request on the same branch, containing only the new updates. 
(This is after the first pull request which contained several other additions, which I submitted previously). 
From what I'm able to see, it looks like this specific one only contains the files related to EfficientGlideRecord. 
Have I done this incorrectly? I'm sure the mistake is mine, but I can't see where it might be. 

Perhaps I just need to sync up my branch with the source, and then re-create my files? 

image

earlduque:
Hey yeah it's a learning process if you don't do too much stuff on git, no worries. So what you're looking at is the commits tab that includes all commits on the source branch of this PR.
image
You're screenshot is specifically looking at the bottom commit, but what trips people up often on GitHub is that a Pull Request is to an entire branch. So the top commit in that picture is also included in this PR since it's on the same branch.

This is usually prevented/solved by ensuring you create a new branch for every separate _feature_ (or in this case, every code-snippet), so that each Pull Request only has the files pertaining to that specific feature/snippet

To see _everything_ your current PR is trying to add/change/delete, check out the `files changed` tab instead: 
image

earlduque:
But I also see that you changed the title of the description to be more inclusive now so technically this is mergeable!

PR #554, opened 2022-10-20
Business rule set program business duration

MrChrisE:
Adding a new business rule that ensures the program duration excludes weekends and public holidays.


MrChrisE:
Adding a new business rule that ensures the program duration excludes weekends and public holidays.

PR #573, opened 2022-10-25
Copy Comments

Radhe-Manasa:
Copy work notes comments as RITM additional Comments.


niamccash (changes requested):
@Radhe-Manasa can you please add these files to a proper subdirectory (eg. Business Rules)?

I'm also noticing that your readme mentions closed states but that's not reflected in the code?

PR #592, opened 2022-10-27
Add standard change model

stevezero:
script to add the standard change model which is not available interactively on an existing change record


niamccash (changes requested):
Hi @stevezero, could you please remove the already contributed UserUtil files from this PR and the errant .vscode directory and contents? Would appreciate the PR clean and organized.

stevezero:
whoops sorry about that. I'll fix it

PR #617, opened 2022-10-30
Updating sp-modal

hafsa-asif:
Add callback in spModal


niamccash (changes requested):
This PR looks incomplete. Please check it again. It's also missing a readme file.

niamccash (changes requested):
Apologies @hafsa-asif I'm just now seeing that there is already an sp-modal directory with some code it in in the repository.

hafsa-asif:
> Apologies @hafsa-asif I'm just now seeing that there is already an sp-modal directory with some code it in in the repository.

Hi. I have actually updated the code of spModal and the readme file with the callback functionality. The existing code did not have the callback functionality which is something very frequently needed.

PR #620, opened 2022-10-30
Azure DevOps Rest Message

hafsa-asif:
Creating a task with the type of issue in Azure from servicenow


niamccash (changes requested):
Please add the readme file

PR #621, opened 2022-10-30
convert space to %20 for url

hafsa-asif:
Converting space to %20 in the string to make it usable for url


niamccash (changes requested):
Also missing a readme file. Please check the code again.

PR #629, opened 2022-10-31
Requested Item variables summary

Radhe-Manasa:
This code help to set the RITM description as the summary of catalog item variables.
we can reuse the code on catalog workflows in run script activity and flow designers


earlduque:
There doesn't appear to be any changes in this Pull Request

PR #656, opened 2023-09-30
Consolidate, document, and refine

Litwilly:
Had noticed one of my contributions from a previous year dealing with array manipulation had a fellow cousin in the Script Include directory so....

- let's consolidate into one array class of utilities
- follow appropriate standards of naming classes (capitalize first letter) 
- make sure we have clear documentation in our readme that includes an example
- let's follow JSDOCs standard for providing function documentation

If this merge is approved I have additional array utilities I would like to include in this class


Lacah (approved):
Good stuff following up on previous commit and consolidating functions following leading practices. Thanks!

PR #658, opened 2023-09-30
Open in Service Ops Workspace UI action

Litwilly:
Simple UI action used in classic view with any task records to open in SOW.

This time I remembered to save!!


Lacah (approved):

Lacah:
Much better indeed, thank you! 😊

PR #663, opened 2023-10-01
Switch case

at8807602:
This is switch statement shows, how you can test the multiple values against the same field..


at8807602:
This is awesome functionality to use with onload client script.

SapphicFire (approved):
Looks good to me!

PR #664, opened 2023-10-01
Create Multiple RITM using single request V.0

vishalbirajdar0712:
Create multiple RITM under single request depending upon list collector values through workflow without using CartAPI & Order guide.

When end user selects the multiple users (let say 3) in “User list” then 3 RITM to be generated under single request.
It should calculate the price of all 3 RITM on portal.


Lacah (changes requested):
Thank you for your contribution. However, it doesn’t comply with our contributing guidelines. As a reminder, the general requirements (as outlined in the [CONTRIBUTING.md file](https://github.com/ServiceNowDevProgram/code-snippets/blob/main/CONTRIBUTING.md)) are the following: follow the folder+subfolder guidelines and include a readme.md file explaining what the code snippet does. In you specific case, your contribution should be please within one of the existing top level categories, e.g. "Server Side", and should not have its own top level folder.
Review your contribution against the guidelines and make the necessary adjustments. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

PR #666, opened 2023-10-01
Verify whether a date falls within a hour range

hoederath:
Adds a code snippet that verifies whether a date falls within a hour range.


Lacah (approved):
I'm approving this, but for future contributions, a little more comments (in the code or in the readme or both) would be helpful. Less experienced people might be having some trouble understanding how to replace certain parameters in the code for their own use.

PR #668, opened 2023-10-01
Feature/copy fields from one table

pandeysumit-now:
# Copy Fields from One Table to Another

This script facilitates copying fields from one table to another. It accepts Table A and Table B as input along with the field names to copy.

## Use-Cases

- This is particularly useful for building demos/POCs on PDI instances when you frequently need to create new tables. In most cases, instead of creating fields one by one, you can copy common fields from one table to another.

**Note:** This is a sample script. Please run this script in a non-production environment first and test for all scenarios. The execution time of this script may vary depending on the number of fields to copy.


Lacah (approved):
Thanks for the proper readme 😊

PR #677, opened 2023-10-01
Decrypt password field

goranlundqvist:
This code snippet can be run as a background script to retrieve a password that you have forgotten and exists in the "basic_auth_credentials" table. It goes in and decrypts the password field and shows what value it has.

I have removed a similar version in the Server Side category since there now is one in the Background Scripts folder where you will probably be running the code snippet. The Code snippet in there is also more documented


Lacah (approved):

Lacah:
Thanks a lot!

PR #681, opened 2023-10-02
Formula Builder Calculate Age from Birthdate

Danmax:
Took me 99 Tries but 100th is the Charm!


earlduque:
That last push is all that counts! Good job, looks great

PR #683, opened 2023-10-02
Regex to check if the email format is valid or not

DhanushNehru:


Lacah:
Thank you for your contribution. Please remember to follow the guidelines in the CONTRIBUTING.md of this repository. ES2021 code must be called out or highlighted, because it is not a default in ServiceNow (ES5 is). You should either mention that or provide an ES5 compatible version as well.
Updated your redme.md for this one time.

PR #688, opened 2023-10-02
Graph ql api snippet

at8807602:
I have created this GraphQL API, easiest way to get the exact data which we want from the server. It overcomes the under fetching and over fetching as compare to REST API.


Lacah (approved):

Lacah:
Great stuff, thanks for the update!

Issue #416, opened 2021-10-30
window == null to determine if in service portal not working


Litwilly:
I've seen a few references to using `window == null` when trying to determine if in the platform or service portal. This only seems to work in my Rome PDI if `isolate script` is set to `false`

https://github.com/ServiceNowDevProgram/code-snippets/blob/b189fc01a77190556a20894fb56896e6ef44a4b7/Client%20Scripts/Get%20URL%20Parameters/script.js#L4-L14

I've found a workaround by using a `try/catch` block like the following:

https://github.com/litwi1rm/code-snippets/blob/63e1a015cecc56611aec8e03523e4fc2f631e555/Catalog%20Client%20Script/Clear%20all%20fields/script.js#L11-L24

Has anybody else ran into this?

abhishek-kumar-91:
I want to work on this issue, can you assign me?

Lacah:
@abhishek-kumar-91 I have assigned it to you, but for your information, we had a pull request already that contained a fix for it, but we were not able to accept that yet. Since then no updates, so if you do it first, the glory is your 😊. Remember to reference this issue in your PR.

rajanarahul93:
"To determine the platform or service portal in Rome PDI with script isolation enabled, consider using try/catch blocks for reliable identification."  ASSIGN ME THIS #416

Lacah:
@abhishek-kumar-91 are you working on this or can I re-assign to @rajanarahul93?

abhishek-kumar-91:
@Lacah yeah, I am still working on this issue.

SapphicFire:
Hi @abhishek-kumar-91 - just following up to see where this is at, and if you'll be able to do this one?

Want to ensure we keep these moving and open it up to the broader group if needed

Lacah:
Open for new contributors to solve this in 2024 too :)

PR #691, opened 2023-10-02
window == null to determine if in service portal not working #416 "SOLVED"

Maithani-ji:
Screenshot (106)

ONLY THIS MUCH IS CHANGED IN THE CODE  --> code-snippets\Client Scripts\Get URL Parameters\script.js

I have fixed the If(windows==null) code by using if(typeof window !== 'undefined' && typeof document !== 'undefined') .
REASON::
/ PROBLEM FACED WITH //if(window==null)
// **1.In a standard browser environment, window is a global object representing the browser window. Checking window == null typically evaluates to false because window exists.
// 2.In a Node.js environment, window is not defined, and checking window == null will usually evaluate to true because window is not defined, and null represents the absence of an object.
// 3.In certain contexts or JavaScript runtimes (e.g., in some server-side environments), window might be defined differently or not at all, leading to varying results when checking window == null.
// 4.he behavior you mentioned, where window == null evaluates to true when the isolate script flag is set to false, could be specific to certain JavaScript environments or configurations. The exact behavior may vary**
Solution::
/ This code will run in non-browser environments (including isolated scripts in PDI)
//if(typeof window !== 'undefined' && typeof document !== 'undefined')


Lacah:
Thank you for your contribution. However, it seems that the file changes in this Pull Request are incorrect or invalid. Your Pull request contains a lot more files than intended, as I believe you only wanted to edit one script to fix issue #416.
Please review the files you have changed/added and ensure that all modifications are correct, valid, and align with the project’s goals and guidelines. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

PR #700, opened 2023-10-02
Call script include

seviceN:
We will have client callable script include function which takes inputs by using this.getParameter('sysparm_xxx')

Sometimes we want to use the same function from server side code, the given code will help us in dealing those situations.


Lacah (approved):

Lacah:
Great stuff, thanks!

PR #702, opened 2023-10-02
Scripted rest api

at8807602:
I have created Scripted REST API, we can easily be able to fetch the group information with the help of this API. If we provide the name of group, it will handle the request, process it and it will return the information like (group email, group manager name etc.)

### DIFF
--- a/Scripted REST Api/Get group info/group_info.js
+++ b/Scripted REST Api/Get group info/group_info.js
@@ -1,0 +1,15 @@
+(function process(/*RESTAPIRequest*/ request, /*RESTAPIResponse*/ response) {
+
+    // implement resource here
+	var group = request.queryParams.Name;  // This will store the provided group name.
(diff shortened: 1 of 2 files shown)


Scripted REST Api/Get group info/group_info.js
```diff
@@ -0,0 +1,15 @@
+(function process(/*RESTAPIRequest*/ request, /*RESTAPIResponse*/ response) {
+
+    // implement resource here
+	var group = request.queryParams.Name;  // This will store the provided group name.
+	var grGroup= new GlideRecord('sys_user_group');
+	var obj = {};  // This will create an empty object.
+	grGroup.addQuery('name',group);  // This will validate, if group exist or not.
+	grGroup.query;
```
earlduque, line 8: incorrect glide method syntax

earlduque (changes requested):

PR #716, opened 2023-10-03
Open modal widget onsubmit

demetris7:
This can be used in an Onsubmit catalog client script to stop submission of the form and open a widget in a modal view. This also uses an async call to fetch some data that may be necessary and based on that open the modal (using SPModal) or continue with the submission.


SapphicFire (approved):
Great contribution, appreciate your input!

PR #727, opened 2023-10-03
Find sys_id named records

iar42:
Background script to find records that contain a sys_id value in a specified field (other than the actual sys_id field)


niamccash:
@iar42 It's likely not that critical and is an edge case but thoughts on using the end of string ($) check in the regex for the sys_id?

iar42:
@niamccash , good catch, thanks! I've added the end of string check in a new commit.

PR #729, opened 2023-10-04
Return user asset util

JonSandoval:
**Code Description**

Client Callable GlideAjax script include to return one of the Following based on passed user's sys_id:

-Single Asset for User

-Single Asset for User with Model Category filtering

-All Assets for User

Script queries the "alm_hardware" table. Intended to be called from a client script/catalog client script, one use case would be to populate a users asset on a catalog item form.

**Files Included**

New folder "Return Asset(s) for User" under "GlideAjax"
Script: "FindUserAsset.js"
Readme file: "readme.md"

**Notes**

Opening a new pull request as my pull request form yesterday did not include a readme.md


SapphicFire (approved):
Nicely done, thanks for contributing. Would suggest supporting providing other details as parameters for functions, such as getAllAssets, just as future food for thought

PR #734, opened 2023-10-04
Avinash238 patch 1

Avinash238:


earlduque:
Instead of deleting another persons way of doing the code. We suggest you upload your code as a separate .js file in the same folder as a separate variant. 

Additionally our contributing guidelines require that pull request titles and descriptions be descriptive to what is being changed. Closing for now.

PR #736, opened 2023-10-05
User role

riya-misra-1:
I have created a custom action for a flow to check the role of user who have submitted the request in catalog item

### DIFF
--- a/Flow Actions/Check User Role/readme.md
+++ b/Flow Actions/Check User Role/readme.md
@@ -1,0 +1,8 @@
+This is a custom action that checks the role of the user who have submitted the request in catalog item
+
+Action input: 
+1. user Type-Reference.User and Mandatory
+2. role Type-string and Mandatory
+
+Action output:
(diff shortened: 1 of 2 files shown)


riya-misra-1:
I have tried it, it was not working in the case when running the flow as a system user.
It is for the case when we want to know the role of user and running the flow as system user. Using gs.getUser()  will give admin role not the actual role of the user.

MartinStoyanoff:
@chelming feel free to take it over

Lacah (changes requested):
@riya-misra-1 I understand your point about running as system user, but I would still encourage you to look at Chris's suggestion regarding the optimization of your query, i.e. including inputs.role as an initial query parameter and limiting your query to 1 result. Thanks.

Lacah:
Closing this for now. Please update your script per the above discussion and reopen this PR or submit a new one, and we'll be happy to approve it! Thanks!

PR #737, opened 2023-10-05
Fetch parameter

riya-misra-1:
I have created catalog client script to fetch the parameter from url and set the field to value of parameter


SapphicFire:
It appears that this contribution is a duplicate or resembles spam, potentially aiming to game the Hacktoberfest event. As a reminder, submissions marked as spam may lead to account restrictions or disqualifications. Please ensure to make unique and meaningful contributions, and refrain from spamming or duplicating submissions. Closing this as spam.

Hi - in this case, this duplicates code that already exists and applies for portal/mobile and platform. As this is not adding a unique approach, I'll need to close this. Appreciate the energy and initiative though, look forward to seeing more from you!

PR #738, opened 2023-10-05
Move customer updates

shivamTodwal:
I've developed a script to facilitate the transfer of customer updates between two update sets within the same application scope. The script needs to run at the global scope for execution.


Lacah (approved):

Lacah:
Thanks for your contribution!

PR #739, opened 2023-10-05
Close child incident

riya-misra-1:
I have created ui actions that close the child incident directly from the parent incident.

### DIFF
--- a/UI Actions/Close child incident/clientScript.js
+++ b/UI Actions/Close child incident/clientScript.js
@@ -1,0 +1,6 @@
+function popUpClientScript(){
+	var answer = confirm('Are you sure you want to close the attached child incident'); // confirm message to ask for confirmation
+	if(answer == true){ //check if answer is true 
+		gsftSubmit(null,g_form.getFormElement(),'sys_demoaction'); // calling server ui action to update state
+	}
+}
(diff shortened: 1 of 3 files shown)


UI Actions/Close child incident/clientScript.js
```diff
@@ -0,0 +1,6 @@
+function popUpClientScript(){
+	var answer = confirm('Are you sure you want to close the attached child incident'); // confirm message to ask for confirmation
```
Lacah, line 2: Please note that using confirm and similar basic Javascript Window methods (such as alert, prompt, etc.) is not recommended in ServiceNow (amongst other limitations, these windows are rendered by user's browser and not by ServiceNow). I would encourage to explore the [GlideModal API ](https://developer.servicenow.com/dev.do#!/reference/api/vancouver/client/c_GlideModalV3API)as a replacement for future use (there are many out of the box UI page to use there, including at least 2 for confirm).
I will accept to code as it is technically correct, but if you'd like, you are welcome to commit an update to it using GlideModal!

Lacah (approved):

Lacah:
As mentioned in line:

Please note that using confirm and similar basic Javascript Window methods (such as alert, prompt, etc.) is not recommended in ServiceNow (amongst other limitations, these windows are rendered by user's browser and not by ServiceNow). I would encourage to explore the [GlideModal API ](https://developer.servicenow.com/dev.do#!/reference/api/vancouver/client/c_GlideModalV3API)as a replacement for future use (there are many out of the box UI page to use there, including at least 2 for confirm).
I will accept to code as it is technically correct, but if you'd like, you are welcome to commit an update to it using GlideModal!

PR #740, opened 2023-10-05
Generate next unique value based on sequence

deepak-64742:
function inside action design to generates a unique sequence value based on the existing sequence in a column of a table.


Lacah (approved):

Lacah:
Thanks for your contribution!

PR #747, opened 2023-10-06
Created Scripted Rest Resource to get data from table return as per p…

dadhich-ashish:
Created Scripted Rest Resource to get data from table return as per parameter pass


Lacah:
Thanks!

PR #750, opened 2023-10-06
Mark an Email High Importance when initiated from an Email Client

sanjivmeher:
This Business Rule can help agent mark an email High importance from Email Client. Out of Box there is no way an agent can mark an email High importance while drafting an email from email client. This Business rule can mark an email High importance when an exclamation (!) is added to the start of the subject in the Email Client

The other scripts 'Add Users in Watchlist to CC' can add users in watchlist of a task to CC in an email.


SapphicFire (approved):
Looks good. In future I recommend ensuring you use .js to syntax highlight and format the code correctly

sanjivmeher:
Thanks. Good Feedbacks

PR #752, opened 2023-10-06
feat(qrcode): generate qrcode

domuserk:
This widget can help developers create a qrcode quickly


domuserk:
Would using external dependencies be a problem?

Lacah:
@domuserk generally no, and there's nothing in our contributing guidelines either to prevent that. It's more a concern of mine in general whether we want to offer code snippets at all that are dependent on a library that is external to ServiceNow. 
I am consulting with my fellow reviewers on that (which may take a bit longer during the weekend), and as such, I appreciate your patience.

Lacah (changes requested):
Thanks again for your contribution! After some deliberation we have decided that it is OK to accept it, however we would like to ask you to expand your readme file a little bit in an effort to clarify the usage of external libraries in this context for less experienced user as well, e.g.:
- steps to take to import a library
- where to include it (e.g. JS includes of a portal?)
- how and from where to call it

It doesn't need to be too long, just please address these points. Many thanks!

Lacah (approved):

Lacah:
Brilliant, thanks a lot!

domuserk:
> Brilliant, thanks a lot!

Thank you for your help

PR #757, opened 2023-10-07
Service portal widget HTML table from GlideRecord

ashoo-jindal:
Create a simple HTML table by fetching the Incident data from server using GlideRecord and pass it using JSON.


SapphicFire (approved):
Looks good to me, thanks for the contribution

PR #760, opened 2023-10-07
Reference lookup form behaviour control

annaydas:
This client-side code snippet allows to control form behaviour when a new record form is opened from the list view displayed by clicking the lookup icon of a reference field.

There might be a requirement to control the form's behaviour when the new record form is opened from a designated field on a specific table.

The provided code snippet can be used in a client script/scripted UI policy defined on the reference table and allows usage of client-side APIs.

**Path to newly added folder: Client Scripts/Control Form Behaviour from Reference Lookup**


Lacah (approved):

Lacah:
Cool stuff, thanks for sharing!

PR #761, opened 2023-10-07
feat(typed_array_elements): add typed array elements snippet

domuserk:
This background script may be useful if you come across a situation where it would be necessary to convert elements to a specific type such as string or number.

### DIFF
--- a/Background Scripts/Typed Array Elements/readme.md
+++ b/Background Scripts/Typed Array Elements/readme.md
@@ -1,5 +1,18 @@
-# Attach Workflow to existing records
+# Type Elements Of An Array
 
 This background script may be useful if you come 
 across a situation where it would be necessary to convert 
 elements to a specific type such as string or number.
+
+# Use Case: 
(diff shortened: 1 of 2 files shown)


Background Scripts/Typed Array Elements/readme.md
```diff
@@ -0,0 +1,5 @@
+# Attach Workflow to existing records
```
Lacah, line 1: What is this line here?

Lacah (changes requested):
The readme file is a bit confusing, especially with the first line. Could you please update it so that it is obvious to less experienced users as well (who may not know how array.map works) what your script is about and how to use it?
Thanks.

Background Scripts/Typed Array Elements/readme.md
domuserk: its a mistake sorry

Lacah (approved):
Awesome, thanks!

PR #765, opened 2023-10-07
Auto_approve approvals

Jaspalsinghchot:
Use this Business rule to get the approval record auto-approved if the same record is approved by approver before. This helps avoid redundant approvals to be actioned by same approver.


earlduque:
I went ahead and made the subfolder name more descriptive

PR #770, opened 2023-10-08
Parse csv file using script and read each row

gowdah:
Parse CSV as a file (not a CSV string input using sn_impex.CSVParser)


earlduque:
Thanks for fixing that!

PR #774, opened 2023-10-08
Search Source - Approvals

Jaspalsinghchot:


Service Portal/Search Sources/DataFetchScript.js
```diff
@@ -0,0 +1,51 @@
+(function(query) {
+    var results = [];
+    var getuserdelegateof = new GlideRecord('sys_user_delegate');
+    getuserdelegateof.addQuery('delegate', gs.getUserID());
+	getuserdelegateof.addEncodedQuery("endsONToday@javascript:gs.beginningOfToday()@javascript:gs.endOfToday()");
```
SapphicFire, line 5: By using 'ON' this code could technically run outside delegation window

I'd suggest something like
`starts<=javascript:gs.endOfCurrentMinute()^ends>=javascript:gs.beginningOfCurrentMinute()`

Service Portal/Search Sources/DataFetchScript.js
```diff
@@ -0,0 +1,51 @@
...
+    getuserdelegateof.query();
+    if (getuserdelegateof.next()) {
```
SapphicFire, line 7: Your query could return multiple results, I'd suggest iterating through them and using a .hasNext() on this initial line

Service Portal/Search Sources/DataFetchScript.js
```diff
@@ -0,0 +1,51 @@
...
+            myapproval.number = grApprovals.sysapproval.number + "";
+            myapproval.short_description = grApprovals.sysapproval.short_description + "";
+            myapproval.sys_created_on = grApprovals.getValue('sys_created_on');
+            myapproval.url = "?id=approvals&table=sysapproval_approver&sys_id=" + grApprovals.getValue("sys_id");
+            myapproval.label = grApprovals.sysapproval.number;
+            myapproval.primary = grApprovals.sysapproval.number;
+            results.push(myapproval);
+        }
+
+        return results;
```
SapphicFire, line 27: This returns before any approvals for the user could be returned. Suggest not returning at this point

Service Portal/Search Sources/DataFetchScript.js
```diff
@@ -0,0 +1,51 @@
...
+    } else {
```
SapphicFire, line 28: Would suggest that the intended behaviour does not require this else. Rather, let the following lines run under all cases, you just want to do your delegate check also :)

SapphicFire (approved):
Hi @Jaspalsinghchot - great contribution. I'll accept this and merge it, but I've added some comments you may want to incorporate into a future version

PR #783, opened 2023-10-09
User difference

riya-misra-1:
I have created scripted rest api that compare two different users and display the difference of field with the values


SapphicFire (approved):
Great PR, thank you

PR #791, opened 2023-10-09
Timestamp extraction from the last journal comment

ivan-betev:
This code snippet demonstrates how to extract a timestamp from a comment text and create a 'GlideDateTime' object from it.
It may be useful if you don't want to drill down to the 'sys_journal_field' table.


Background Scripts/Get The Last Journal Comment Date/README.md
Lacah, line 1: @ivan-betev thanks for this cool and unique contribution. Perhaps it would make sense to mention the specific format this regex can match, i.e. DD-MM-YYYY HH:MM:SS?

Lacah (changes requested):
Please update readme with some more details around what is being matched (user readable format). Thanks.

Lacah (approved):

Lacah:
Thank you so much for the quick update. Cool commit!

Issue #745, opened 2023-10-06
Replace standard Window methods with GlideModal equivalent


Lacah:
It is not recommended to use the standard Javascript Window methods for pop-ups. Let's get them replaced with GlideModal (or spModal for portal), i.e.:

**Confirm**
- [x] [Copy Variable Set UI Action](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/UI%20Actions/Copy%20Variable%20Set/scripts.js#L3)
- [x] [Close Child Incident UI Action](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/UI%20Actions/Close%20child%20incident/clientScript.js#L2)

**Alert**
- [x] [Mandatory Attachments Client Script](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/Catalog%20Client%20Script/Mandatory%20Attachments%20with%20'n'%20numbers/onSubmitClientScript.js#L8)
- [x] [Block Submit Catalog Client Script](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/Catalog%20Client%20Script/Block%20Submit/block_submit.js#L5)
- [x] [Mark Records Inactive UI Action](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/UI%20Actions/Mark%20Records%20Inactive%20-%20List%20Action/listAction.js#L17)
- [x] [Currency Validation Catalog Client Script](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/Catalog%20Client%20Script/Currency%20Validation/currency_validation.js)
- [x] [GetCallerDetails Script Include client caller script](https://github.com/ServiceNowDevProgram/code-snippets/blob/e2bcf76588fd4fe7c4a625c08a918132f37747ef/Script%20Includes/GetCallerDetails/Calling%20Script%20Include%20from%20client.js#L11)

Found more? Feel free to add to the list!

Had a merged PR that fixes one or more of these? Tick 'em off!

kmohit2809:
[...]

PR #792, opened 2023-10-09
Fix/745 block submit catalog client script

sheikhafsarzap:
Fix issue #745 
To complete task "Block Submit catalog client script".
Updated Readme.MD and script in block_submit.js


Lacah:
Thank you very much!

PR #799, opened 2023-10-09
RM: New stopwatch script include

RahmanM:
RM: New stopwatch script include


MartinStoyanoff (approved):
Looks good, with the very minor remark that it is not a good practice to use gs.sleep().

PR #801, opened 2023-10-10
Clone user along with roles and groups.

shivamTodwal:
I have created a script in ServiceNow to replicate a user's profile. This script not only duplicates the user's data but also replicates the roles and groups assigned to that user.


Lacah:

PR #802, opened 2023-10-10
Add rows in mrvs

goranlundqvist:
Not much to add, a simple catalog client script to push data into the MRVS


kmohit2809:
Good one @goranlundqvist. Thanks for your contribution.

PR #803, opened 2023-10-10
Read workspace url

Iuliak:
Display Business Rule used to read a parameter from the sow or agent workspace URL. In this BR a parameter caller_id is read, which contains an email address. The email address is then used to find the appropriate user in the sys_user table, which then will be set in a scratchpad variable. This can be used for eg in an onLoad Client script to set the default value of a field.

image


Lacah (approved):

Lacah:
Thanks for updating and contributing!

PR #809, opened 2023-10-11
Get my reports dynamic filters

smartcarleen:
This looping script traverses the User table from a certain point to get either one level of employees or all employees in the hierarchy underneath the logged-on user. There are two functions:
1. **getMyDirectReports**: gets only users directly reporting to the logged on user
1. **getMyReports**: gets all users reporting to the logged on user

This solution has three components: one Global Business rule and two Dynamic Filters.
* Admins can use the script as a Reference Qualifier
* End Users can select the predefined filter in lists and reports (like with "One of My Assignments").


Lacah (approved):

Lacah:
Huh, had no dynamic filters folder until now - I guess it makes sense! Goodie!

PR #810, opened 2023-10-11
Client script multi value ajax call without JSON

abhrajyotikanrar:
This code snippet allows us to receive multiple responses from the server side to the client side using GlideAjax without using the complex JSON format. You can set multiple attributes on the server side, and use the same attribute on the client side to receive responses.


abhrajyotikanrar:
This code snippet allows us to receive multiple responses from the server side to the client side using GlideAjax without using the complex JSON format. You can set multiple attributes on the server side, and use the same attribute on the client side to receive responses.

Lacah (approved):

Lacah:
Interesting, thanks for your contribution!

PR #815, opened 2023-10-11
Scripted REST API to insert Tags.

snamjosh:
This is utility which contains scripted REST API to insert tags in servicenow records.


Lacah:
Thank you!

PR #817, opened 2023-10-12
feat: new function returns variables as string

ronaldkarim:
variablesToText checks record variables & returns Question & answer to string.
Returns all Variables with an answer as a formatted string. 
Can be used to apply question:answers to string fields or notifications

Added to CatalogUtils,this Script include has similar functionality.
Included example use and example output as per existing template.

Formatting updates to Script Include.


Lacah (approved):

Lacah:
Great, thanks!

PR #818, opened 2023-10-12
Assign role custom flow action

shivamTodwal:
I have developed a custom action that assigns a specific role to a user. This action takes the username and role as inputs. The role input is of reference type.

### DIFF
--- a/Flow Actions/Assign Role/assignRole.js
+++ b/Flow Actions/Assign Role/assignRole.js
@@ -1,0 +1,20 @@
+(function execute(inputs, outputs) {
+
+    var user = inputs.user_id;
+    var roleSysId = inputs.role_sys_id;
+
+    var userGR = new GlideRecord("sys_user");
+    userGR.addQuery('user_name', user);
+    userGR.query();
+
+    if (userGR.next()) {
+        var userSysId = userGR.getValue('sys_id');
+    }
+
+    var userRole = new GlideRecord('sys_user_has_role');
+    userRole.initialize();
+    userRole.setValue('user', userSysId);
+    userRole.setValue('role', roleSysId);
+    userRole.insert();
+
+})(inputs, outputs);
--- a/Flow Actions/Assign Role/readme.md
+++ b/Flow Actions/Assign Role/readme.md
@@ -1,0 +1,8 @@
+# Assign Role
+I have developed a custom action that assigns a specific role to a user. This action takes the username and role as inputs. The role input is of reference type.
+
+**Action Input**
+![action-input](https://github.com/shivamTodwal/code-snippets/assets/146908877/c636f0b7-843c-4808-97ab-e9318e3485d9)
+
+**Script Input**
+![script-input](https://github.com/shivamTodwal/code-snippets/assets/146908877/20f0532f-ad5e-41d8-abe0-4c39ba05b123)


Flow Actions/Assign Role/assignRole.js
```diff
@@ -0,0 +1,20 @@
+(function execute(inputs, outputs) {
+
+    var user = inputs.user_id;
+    var roleSysId = inputs.role_sys_id;
+
+    var userGR = new GlideRecord("sys_user");
+    userGR.addQuery('user_name', user);
```
Lacah, line 7: This GlideRecord query could be optimised:
```
var userGR = new GlideRecord("sys_user");
userGR.get('user_name',user);

if(userGR) {
```

Flow Actions/Assign Role/assignRole.js
```diff
@@ -0,0 +1,20 @@
+(function execute(inputs, outputs) {
+
+    var user = inputs.user_id;
+    var roleSysId = inputs.role_sys_id;
```
Lacah, line 4: I also don't see the value in using the sys_id of the role. It would be much better to just take the role name, such as `user_admin`.

Flow Actions/Assign Role/assignRole.js
Lacah: Or if you decide to use user sys_id as input, it should simply be `userGR.get(user)`.

Flow Actions/Assign Role/assignRole.js
```diff
@@ -0,0 +1,20 @@
...
+    userGR.addQuery('user_name', user);
+    userGR.query();
+
+    if (userGR.next()) {
+        var userSysId = userGR.getValue('sys_id');
+    }
+
+    var userRole = new GlideRecord('sys_user_has_role');
+    userRole.initialize();
+    userRole.setValue('user', userSysId);
```
Lacah, line 16: You don't have a user sys_id. You are asking for user_id inn inputs. You should decide which value you want to to use and update your script to be consistent.

Flow Actions/Assign Role/assignRole.js
```diff
@@ -0,0 +1,20 @@
...
+    userRole.setValue('role', roleSysId);
```
Lacah, line 17: As mentioned above, decide if role sys_id really makes sense or not.

Lacah (changes requested):
Please review the comments provided and update your code block accordingly.

Lacah:
Thanks for your contribution. I have provided some comments on your code block to optimise it a little bit. Please review, update, and test your code.
I will close this pull request now, but once you are done with your updates, please feel free to reopen it, or create a new one.
Many thanks.

Flow Actions/Assign Role/assignRole.js
shivamTodwal: Hi Lacah, As to assign a role to any user we need to create a record in "sys_user_has_role" table for that user. This table contains a role field, and user field. As both fields are reference fields, so we need to use the sys_id of the role.

Flow Actions/Assign Role/assignRole.js
shivamTodwal: Hi Lacah, As to assign a role to a particular user we need to create a record for that user in the "sys_user_has_role" table. In this table the user and role field are of reference type, and to set the value of reference field we need sys_id. This is the reason why I am using the sys_id for the role and user. In the input I am taking the user name and the role as input. I have attached the screenshot for the same in the readme.md file.

shivamTodwal: Thanks Lacah, for you valuable suggesstion. I will update this and will raise the new PR for this.

PR #819, opened 2023-10-12
Confirm box using glide modal

kmohit2809:
This is a sample script to generate a confirm box with 'Don't save' , 'Cancel' and 'Save' buttons. This is created using GlideModal instead of standard Javascript window methods which can be easily blocked by user's browser.


Lacah (changes requested):
Thanks Mohit, but could we put it into one of the existing folders, either Client Scripts or UI Actions instead of creating a new one for GlideModal?
Thanks.

kmohit2809:
Hi Laszlo, sure we can do that, but I thought of putting few more OOB Ui pages for glidemodal hence put it in a new folder. Please let me know.

Lacah:
OK sure buddy, might make sense in this case. Thanks!

Lacah (approved):

PR #820, opened 2023-10-12
Preview context record during approval

abhrajyotikanrar:
While approving any request it was very hard until now to preview the record for which the approval was required. This UI action created on sysapproval_approver table will enable previewing the record before approval so that the approver can make an easy informed decision.


abhrajyotikanrar:
@Lacah @earlduque Could you please review and approve this pull request if possible?

PR #821, opened 2023-10-12
ServiceNow - ISO8601 Date Parser

ivan-betev:
Some APIs commonly return dates in ISO8601 format. This script may help.


Lacah (approved):

Lacah:
This is cool, thanks!
Idea for another PR: I actually needed the revers of this too a few times for outbound APIs 😉 Flow Designer has a built in transformer for that, but it may not be an option for every use case.

Thanks again, great contribution!

PR #825, opened 2023-10-13
Autopopulate email

riya-misra-1:
I have created a catalog client script that works on onChange and automatically change the email in the email field according to the change in user in the requester name field.


Lacah:
Thank you for your contribution. Could you please explain what is the context in which script is useful or necessary?
Since the Utah release, ServiceNow provides an 'Auto-populate' option for catalog variables, which can do the same thing without any code. I am not sure in what other use case would this be needed, when the information is always available to the system directly via the user reference. But perhaps you have a convincing argument? Let me know! 😊

riya-misra-1:
Ok Thanks Got it. Actually, I am using Tokyo.

Lacah:
Closing this for now. If you would like to update your contribution, you can either reopen this PR or submit a new one.

PR #832, opened 2023-10-14
Flow action get value from choice known error workaround

soerenmaucher:
Adding a generic Flow Designer Action to mitigate a known error. Currently, choice fields cannot be set properly from the data pill since the runtime value used by Flow Designer is the display value and not the actual value. As a workaround, I propose a simple generic Flow Designer Action, that returns the value of any given choice field which can then be used in Flow Designer.


SapphicFire (approved):
Looks good - in future, queried like this one could use the `.get()` GlideRecord method for slightly more compact code. Good work

soerenmaucher:
Good point. Thanks for accepting.

PR #837, opened 2023-10-14
Hackfest23

madhu1256:


Business Rules/Create Auto-Assign Incident to Support Group/Auto-Assign Incident to Support Group.js
```diff
@@ -0,0 +1,14 @@
+// Situation: When a new incident is created, automatically assign it to the appropriate support group based on the categorization.
+
+(function executeRule(current, previous /*, g*/) {
+    if (current.category == 'Hardware' && current.subcategory == 'Laptop') {
+        current.assignment_group = 'Hardware Support';
```
SapphicFire, line 5: Dot-walking to get and set values is not recommended and will not always behave as expected. Use `current.getValue(field)` and `current.setValue(field,value)` to appropriately access and set these values.

Additionally, setting assignment group by it's display field (name) is also not best practice.

Business Rules/Create Auto-Assign Incident to Support Group/readme.md
```diff
@@ -0,0 +1,5 @@
+Explanation: 
+
+This business rule is triggered when a new incident is created. It checks
```
SapphicFire, line 3: This may be better suited as an Assignment Rule example than a Business Rule

SapphicFire (commented):
Thanks for this submission. I have provided some feedback and request that you incorporate this before re-opening this PR or submitting a new PR.

PR #845, opened 2023-10-14
Fix security issues

ovarn:
The updated client-callable Script Includes are vulnerable and allow an attacker to bypass the ACLs. We should avoid such examples in the repository since not all developers understand the consequences of using "generic" `GlideAjax` for retrieving data using the `GlideRecord`.


SapphicFire (approved):
Thanks for this contribution

PR #849, opened 2023-10-14
Regular expression for Indian mobile numbers

nthumma:


SapphicFire (approved):
Thanks for your contribution - consider providing explanations for parts of the regex in future - for example, clarifying why [789] is documented specifically alongside a 9 digit number, or alternatively an 11 digit number is permitted by the phone system may help folks in using and troubleshooting this.

PR #851, opened 2023-10-15
Code snippets

ynr-ram:
This branch contains 4 Code snippets.

1.  Notifications to managers when critical incidents are open for more than a specified period.
2. Code Snippet of RestMessageAPI using try-catch block for parsing the response and creating the records in ServiceNow.
3.  Count the number of Active Incidents by Assignment Group using GlideQuery.
4. Automate calculates the Due Dates based on priority and Assignment group working hours.


ynr-ram:
5.Calculate the time difference between incident creation and change implementation
Merging with other 4 codes

PR #858, opened 2023-10-15
Add more tables to code search

mskoddow:
This is the initial pull request for a background script that extends the OOTB code search base with all table/field candidates that can hold JavaScript code. That way the code search will also find JavaScript that is contained in condition fields, in (sub)flows, flow actions or even in JSON payloads.


Lacah (approved):

Lacah:
Cool, danke sehr Maik! 🙌

PR #860, opened 2023-10-15
Rest message json parser

ynr-ram:
Rest Message JSON Parser


Lacah:
Thank you for your contribution. While I appreciate that you were trying to illustrate how to parse a JSON response and map to data in ServiceNow, there are a couple of issues due to which it cannot be accepted, i.e.:
- There are numerous examples of parsing a JSON response in the community, developer site, docs, YouTube etc. I do not see anything on top of those here that would add more value.
- This is not a Scripted REST API, just a server side (or background if you like) script.

Due to the above reasons I will now close this PR. If you believe there is in fact a reason for this code to be merged into this repository, please highlight the added value in your description and put it in the correct folder.

PR #861, opened 2023-10-15
Enable vip users

Paulsylo25:
@SapphicFire - Thanks for pointing this out. edited the code. looks slike spell check s has corrected the "g_user.hasrole" to "g_user.hassle". now corrected


Client Scripts/Enable VIP and Senior VIP Checkboxes and read only/ScriptToEnableVIP_superVIP users
```diff
@@ -0,0 +1,11 @@
+function onLoad() {
+   //Type appropriate comment here, and begin script below
+   var myRole= g_user.hasRole("admin");
+	if(myRole!=true){
+		g_form.setReadOnly("vip",true);
+		g_form.setReadOnly("u_senior_vip",true);
```
Lacah, line 6: `u_senior_vip` is not an out of the box field and there the code will generate errors for most users. I recommend removing it and relying on standard fields only.

Client Scripts/Enable VIP and Senior VIP Checkboxes and read only/ScriptToEnableVIP_superVIP users
```diff
@@ -0,0 +1,11 @@
...
+	}else{
+		g_form.setReadOnly("vip",false);
+		g_form.setReadOnly("u_senior_vip",false);
```
Lacah, line 9: ´u_senior_vip´ is not an out of the box field and there the code will generate errors for most users. I recommend removing it and relying on standard fields only.

Client Scripts/Enable VIP and Senior VIP Checkboxes and read only/Readme.md
Lacah, line 1: As I mentioned in the code too, the Senior VIP field does not exists out of the box and as such, should not be referenced. Furthermore, you mention in the readme that the script makes these fields visible, but it is just setting the readOnly attribute. Either update the readme or add the missing code.

Lacah (changes requested):
Please review and correct the issues mentioned. Thanks.

Lacah:
Thank you for your contribution. Due to the issues listed for the contributed files, I will close this PR now, however please feel free to reopen this or submit a new after the corrections were made. Thank you for your understanding.

PR #864, opened 2023-10-16
Autoclose incident

riya-misra-1:
I have created a schedule jobs that will automatically close all the incident that has been resolved last week.

### DIFF
--- a/Scheduled Jobs/Auto close incident weekly/readme.md
+++ b/Scheduled Jobs/Auto close incident weekly/readme.md
@@ -1,0 +1,1 @@
+This is a scheduled job that will check on every monday weekly that last week what all incidents has been resolved. It will close all the incident record on every week monday that has been resolved last week.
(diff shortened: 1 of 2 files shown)


Lacah:
@riya-misra-1 I'm sorry but I have to close this one as well, as it is basically a duplicate of an (old) out of the box feature (and scheduled job), see [here](https://docs.servicenow.com/bundle/tokyo-it-service-management/page/product/incident-management/task/configure-incident-auto-close.html).
From a process perspective it would be quite a bad practise to close incidents resolved last week simply by a weekly job. That would mean some incidents stay inn the resolved state for up to 7 days while others less than a day. That is not a consistent process and quite a problematic user experience. This is exactly why ServiceNow provides a scheduled job out of the box that runs every hour (´Autoclose incidents´), and it is strongly recommended for most customer to use that.

PR #866, opened 2023-10-16
Drag and Drop Widget

chris-canto-11:
This widget allows user to drag and drop objects through the Service Portal page. This demo just provide a way to move and assign/map an object from one container to another side.

The core benefit of this widget is the drag and drop functionality that provide the user a way to dinamically and interactively move object in the Service Portal.

Starting from this demo, it's possible to build even more interesting features for your clients.


Lacah (approved):

Lacah:
Great stuff, thanks a lot for your contribution!

PR #875, opened 2023-10-16
Warn for changed OOTB artifacts

mskoddow:
This is the initial commit for a Business Rule, which displays a warning when opening a modified OOTB artifact. This is helpful to raise the attention on that important topic, as changed OOTB artifacts will created skipped records during the next upgrade.


niamccash:
Thanks for contributing. Looks great.

PR #880, opened 2023-10-17
Look up Support group of CI

NBaker137:
An action to return the **Support group** of a given CI or return a default group (specified as an input) if there is no **Support group** value on the specified CI.


SapphicFire (approved):
Thank you for the contribution

PR #884, opened 2023-10-17
Select a random user from group

rene-el:
A code snippet that can be used to select a random user from a given group.


SapphicFire (approved):
Thanks for the contribution!

PR #885, opened 2023-10-17
Add remove group member

shivamTodwal:
I've developed a script include that facilitates the addition and removal of members from a group.


Lacah:
Thank you for your contribution. One potential improvement idea: addMember() function could check if the provided user - group relationship already exists. Otherwise it's al fine and therefore accepted.

shivamTodwal:
> Thank you for your contribution. One potential improvement idea: addMember() function could check if the provided user - group relationship already exists. Otherwise it's al fine and therefore accepted.

Thanks for the valuable suggestion.

PR #886, opened 2023-10-17
Real time log watcher

abhrajyotikanrar:
This code snippet will give you a new experience of how you see logs while debugging. This UI Page can be navigated by `https://<instance_name>.service-now.com/log_watcher.do` and will give a real-time log capture which will help to make debugging easy for developers. You can also search with the keyword to find your desired log.


Lacah:
Nice one, thanks for your contribution.

PR #889, opened 2023-10-17
Record activity collector

manrick:
Hi Team,

This pull request contains a simple solution how the ticket activities can be collected with the same format, as it can be seen on the record form layout.
This solution is coming from a real project. I had to collect all activities from a record a provide them to the customer as a PDF document.


SapphicFire (approved):
Looks great, thanks for the contribution!

PR #890, opened 2023-10-17
Custom table usage

ajcooper72:
This script provides a way of counting where any custom tables (u_) are used in the instance.


paigesndev:
Hello!

I would recommend adding a disclaimer in your readme that this number may not accurately reflect your total custom table usage. Since it only accounts for tables in the global scope (u_) and doesn't account for tables that extended tables that are exempt from licensing charges individuals using your script may find that their numbers vary significantly from ServiceNow's numbers.

ajcooper72:
Hi @paigesndev 
The purpose of the script wasn't to cover licensing or subscriptions, but more to give an idea of where the table was being referenced from. It's an updated version of a script I wrote a while ago.  I'll add a message to clarify that.

paigesndev:
@ajcooper72 Sounds good! Thanks for the clarification.

PR #893, opened 2023-10-18
Get my groups

manrick:
Hi Team,

The GlideSystem User object has a useful function, which is called getMyGroups. This gives back the sys_ids of current user's group. But the functionality behaves differently depending where it is called from Global or Scoped App.

My solution is a simple function which checks that the group list is a JavaObject or not. If so, it is converted to JS Array. From this point it doesn't matter where it is called from, the result will be the same type of object in both cases.


SapphicFire (approved):
Looks good to me, thanks for the contribution

PR #894, opened 2023-10-18
Background Script to replace HTML tags using script to get the actual text

sonamtiwari8:
This script can be utilized to remove HTML tags from an input string using regular expressions. It replaces all HTML tags with an empty string, leaving only the plain text content. This is useful for extracting text from HTML strings.


SapphicFire:
Thanks for your contribution. This is not quite in a state where we can accept it - the code snippet file needs to be a .js file, and the regex needs to be simplified to not apply over multiple lines or sufficient justification provided. Right now it would match on and remove the valid content below, whether multi line or not.

```When the counter is < 5, that means we will be completing soon.

Otherwise, when the counter is > 20, we need not be concerned```

I'll close this for now, but please reopen or resubmit once you've made the changes

PR #898, opened 2023-10-19
Service portal widget progress bar

debendu-das:


Lacah:
Thank you for your contribution. Since this is not your development but rather the illustration of the usage of an [existing ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role), at the very least I would expect a detailed description and a readme file that is helpful to users and explains how to actually use it in their portals.
I will close this PR for now, but feel free to reopen it or create a new one once you've made updates.

PR #899, opened 2023-10-19
Urgency high

riya-misra-1:
I have created a client script that update the urgency to high when the caller is changed from current to particular caller whose vip is true


Client Scripts/Set Urgency to High onChange of caller field/readme.md
Lacah, line 1: I will let this one go now, but next, please articulate better the context of your script, e.g. that this should be used on the incident table as an onChange script with the Caller field set. It may be trivial for most people, but not for everyone. Thanks.

Lacah (approved):

PR #900, opened 2023-10-20
Category field

riya-misra-1:
I have created a client script that search for particular word in short description at a particular postion and then create one more choice in category field and set the value of category with the newly created field when word found.


Lacah:
Thank you for your contribution. As it is right now, this script is not very useful for the wider community as it seems to be very specific to 1 scenario from your personal work. If you would like this to be included inn this repository, I recommend you make your script more generic, i.e. instead of having hard-coded values, user variables, so users can choose what they want to search for and what value they want to set. Once done, please feel free to reopen this PR or submit a new one. Closing this for now.

PR #905, opened 2023-10-20
Client script to restrict creation of p1, p2 priority incidents

Abhishekchowdarapu:


Attachments/Delete RITM Attachment/deleteattachment.js
```diff
@@ -0,0 +1,15 @@
+// To automatically delete RITM attachment, when an attachment is deleted on SCTASK.
+// Create a After Business Rule, and set business rule condition as Table name is "sctask", so that when an attachment is deleted on SCTASK, this business rule will run.
+
+var gr = new GlideRecord("sys_attachment");
+    var task = new GlideRecord('sc_task');
+    if (task.get(current.table_sys_id)) {
+        gr.addEncodedQuery("table_name=sc_req_item^table_sys_id=" + task.request_item.sys_id + "^file_nameSTARTSWITH" + current.file_name);
```
Lacah, line 7: Is the `file_nameSTARTSWITH` condition a good idea? It is common practise for users the name different attachment with the same name followed by a number, e.g. for screenshots. If, for instance, image.png is deleted on sc_task, and there are other files like image_1.png, image_2.png etc, this would delete all files. I suppose this is not intended behaviour, so I would recommend updating it.
Since in that case you will be searching for 1 file only, you could add `.setLimit(1)` to your GlideRecord to further optimise the database query.
While you're at it, you might want to consider giving a more descriptive name to your GlideRecord variable instead of gr (per leading practise).

Attachments/Delete RITM Attachment/readme.md
Lacah, line 1: Please mention that this business rule should actually be created on the `sys_attachment` table. It may not be obvious for everyone and the comment in the code just makes it more confusing for less experienced people.

Client Scripts/Client Script To Restrict Creation of P1, P2 Incidents/script.js
Lacah, line 1: Please delete this file, it seems to be a duplicate with a very long folder name anyway (and missing readme).

Lacah (changes requested):

Lacah:
Please complete the marked changes and reopen this pull request or submit a new one. Closing this for now.
Thanks for your understanding.

PR #907, opened 2023-10-20
ScriptInclude to convert RITM Variables to Description

anveshmupparaju:
This script include can be used to get all of the variables from an RITM as text. It is client callable, so it can be used from any client script as well as Server-Side script like a business rule.


Lacah (approved):

Lacah:
Thanks!

PR #910, opened 2023-10-20
Business rule for copying attachments from Approval Record to Change Record

anveshmupparaju:
Copy attachments from Approval Record to Change record upon Approval / Rejection


anveshmupparaju:
Hi @kmohit2809 thanks for taking the time to review this.
The business value is, that sometimes approvers attach emails or other documents to justify their decision either to approve or reject.
This BR will copy those attachments to the actual change record which makes it more visible for the change requestor.

This use case has been asked by many of the clients I worked with.

kmohit2809:
Thanks for elaborating.

PR #911, opened 2023-10-20
Password Strength Checker

yuvraj1107thapa:


Regular Expressions/Password Strength Checker/README.md
Lacah, line 1: Thank you for your contribution. It's all fine, but since you are using ES2021, please add a note about this code only working in scoped applications where it is enabled (default for new scopes since Utah). Thanks.

Lacah (changes requested):

Regular Expressions/Password Strength Checker/README.md
yuvraj1107thapa: Got it, Lacah! I've added a note about the ES2021 code to the README. Appreciate the feedback!

PR #922, opened 2023-10-21
Business rule to find implementation status

Abhishekchowdarapu:
Hi @Lacah, @SapphicFire 

Thank you for your valuable feedback!!.

As per your suggestion, I have synced the fork and removed all the unnecessary files, and also updated the code to just trigger an event instead of using a custom field.

Please let me know if any further changes required.


Lacah:
Great job, thanks!

PR #924, opened 2023-10-21
Rest message to automatically create a group in target instance when created in source instance

Abhishekchowdarapu:


Lacah:
Thank you for your submission. However, it appears that this pull request is incomplete. It lacks details about the actual endpoint to use, and as such it is not much use for less experienced users (while more experienced users probably won't need this). Please change your code snippet to explicitly call a ServiceNow API (and method), and don't just refer a REST Message record that only exists in your instance, otherwise this is not helpful to anyone.
Please complete your contribution by adding the necessary elements and resubmit or update this pull request for further review. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

PR #925, opened 2023-10-21
Script Include to sort array of Date strings or GlideDate objects

anveshmupparaju:
A collection of scripts related to sorting an array of Date strings (yyyy-MM-dd format) as well as GlideDate objects.

The business value of this Script Include is that, as a developer one needs to handle the scenarios where the dates need to be sorted, for example in the case of building a custom report. This Script Include will increase the productivity of a developer a little.


Lacah (approved):

Lacah:
Great stuff, thanks for the good description too!

PR #926, opened 2023-10-21
Rest message to automatically create a group in target instance when created in source instance

Abhishekchowdarapu:
Hi @Lacah 

Thanks for reviewing my previous pull request.

As per your suggestion, I have modified the code to utilize the endpoint of the target instance instead of using the Outbound Rest Message.

Please check and let me know if anything else is required to update.

Thanks,


Lacah (approved):

Lacah:
Thank you!!!

PR #928, opened 2023-10-22
Reduce syslog query

mskoddow:
This is the initial pull request for a business rule that protects the user from having their session locked for several minutes due to millions of records at syslog table and opening that table just by entering "syslog.list" into the application navigator.


Lacah (approved):

Lacah:
Smart! Thanks for committing.

PR #930, opened 2023-10-22
Get formatted date

Abhishekchowdarapu:


Lacah:
Thank you for your contribution, however I cannot accept this pull request, due to:

- Missing description after multiple reminders
- The Script Include does not account for different date formats, and therefore it may not be useful in instance where the format is different. It should ideally accept date format (e.g. "dd-mm-yyyy") as a parameter.
- Furthermore, if you alway expect the initial string to be in a single format (such as "yyyymmdd" then the insertion of hypens can be achieved in a single line of code using replace with regex.

See below for an illustration of these:
```
var strDate = "20210421";
var properDate = new GlideDate(); 
properDate.setValue(strDate.replace(/(\d{4})(\d{2})(\d{2})/, "$1-$2-$3"));
gs.info(properDate.getValue()); // => "2021-04-21"
gs.info(typeof properDate); // => object
gs.info(properDate.getByFormat("dd-MM-yyyy")); // => "21-04-2021"
```

PR #932, opened 2023-10-22
Hide variables on catalog for when using cascade on order guide

Abhishekchowdarapu:
This pull request contains a catalog client script that can be used to hide the catalog item variables on the catalog form, when the catalog form is utilized through an Order Guide. So that it can directly use the field values of Order Guide using the cascade feature available on Order guide.

This will eliminated the need to re enter the field values which are already present on the Order Guide Form.


Abhishekchowdarapu:
Hi @Lacah 

Thanks for your valuable feedback!!.

I have added the description in the pull request now. Apologies for the previous pull request.

Also, I have verified that my code is working fine. It is hiding the variables on the catalog form, when the catalog form is opened  on Order Guide.

Could you please let me know where exactly my code is written wrong.

Lacah:
Thank you, I see you have updated the code so you already found the error :)

PR #934, opened 2023-10-22
Assignment restriction

SrijanPatwa:


Business Rules/Assignment Restriction/restrictAssignment.js
Lacah, line 1: Hardcoded sys_ids in scripts represent both a bad practise and a security issue as also mentioned in t[his support article](https://support.servicenow.com/kb?id=kb_article_view&sysparm_article=KB0785129).
This script could be valuable if the restricted (or allowed) group would be defined in a system property instead, per the users requirement.

Lacah (changes requested):

Lacah:
Please review the comments regarding the script. If you reopen this PR or submit a new one, please remember to add a description according to the contributing guidelines. Thanks.

PR #936, opened 2023-10-22
Validate email on import

SrijanPatwa:
Validate the email address when using the OOB 'Import' feature from the list context menu

### DIFF
--- a/Business Rules/Validate Email on Import/readme.md
+++ b/Business Rules/Validate Email on Import/readme.md
@@ -1,0 +1,10 @@
+This business rule can be used to validate the email address when 
+we are using the OOB 'Import' feature from the list context menu to upload some data from spreadsheet into the table.
+
+
+Business Rule will have the below configuration:
+When to run: before insert + update
+Filter condition: Email[u_email_address] is not empty
+
+In the advance >script section paste the code, it will run when we will import any spreadsheet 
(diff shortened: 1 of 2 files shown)


Business Rules/Validate Email on Import/readme.md
Lacah, line 1: Can you please clarify where (on which table) this business rule should be added to? I assume on some user import table? You might also want to clarify that the `u_email_address` field might be called differently, so perhaps replacing it with a variable in the code would be more fortunate.

Lacah (changes requested):

Business Rules/Validate Email on Import/readme.md
SrijanPatwa: This rule can be added to any table which is having email-type field. For example [sys_user] table or any custom table which is used to store email with respect to some username and other details.

[u_email_address] This field is just an example of the email field, it can vary as per the field name in which the user wants to apply this business rule.

SrijanPatwa: I have updated the readme file with more details.

SrijanPatwa: I have also updated the script.

Lacah (approved):

Lacah:
Thanks for the updates!

PR #937, opened 2023-10-23
Update worknotes

riya-misra-1:
i have created a business rule that check if user update work notes of problem task then it should update the same worknotes in parent problem record as well with highlighting problem task number.


Business Rules/Update worknotes of Ptask to parent Problem record/readme.md
Lacah, line 1: Please also mention that this business rule is expected to be created on the problem task table as an after / async business rule.

Lacah (changes requested):

Lacah:
Thank you for submission. Please review the comments added and update your files accordingly, then feel free to reopen this PR or submit a new one. Closing this for now.

PR #938, opened 2023-10-23
Update worknotes

riya-misra-1:
I have updated the business rule which runs after update and on condition when worknote changes,
This business rule update work notes of problem task then it should update the same worknotes in parent problem record as well with highlighting problem task number.


Lacah (approved):

Lacah:
Thanks for the updates!

PR #943, opened 2023-10-24
populating short description with current logged in user

chetnadev:
populating short description with current logged in user and incident number


chetnadev:
hey there, I am a begineer so if there is any issue just let me know and please don't mark this as spam as i really wanna learn and to contribute in this fest. Thankyou

PR #956, opened 2023-10-26
Bulk cat item image update

captainbraddles:
This script allows a user to update the picture and/or icon fields in catalog items in bulk. This can be used for scenarios where the same icon needs to be used for multiple existing catalog items, eg. all Apple related catalog items need to have their icons updated to a different colour Apple icon.


captainbraddles:
Thanks Lacah. I'm still new to github and submitting requests so learning as I go. I'll make the changes and create a new pull request.

Cheers,

Brad

PR #964, opened 2023-10-26
Calculate time difference using GlideDateTime

chetnadev:
This code calculates the time difference in seconds between a fixed date and the current date and time, then logs the result using GlideDateTime


Lacah:
Thank you for your submission. Unfortunately we already have a more versatile and dynamic code snippet that also covers your use case: [GlideDateTime/Get Date Difference](https://github.com/ServiceNowDevProgram/code-snippets/tree/main/GlideDateTime/Get%20Date%20Difference).
As such, I am closing this as duplicate.

PR #965, opened 2023-10-26
Background script for inserting a new record

chetnadev:
the code will create a new GlideRecord object for the 'sys_user' table, initializing it, setting the values for the 'user_name,' 'first_name,' 'last_name,' and 'email' fields, and then inserting a new record into the 'sys_user' table with the specified values.


kmohit2809:
@chetnadev thanks for your contribution, but the script needs a correction, it should be gr.initialize() and not gr.initialise(). Please modify those changes.

chetnadev:
Yes I updated it

chetnadev (commented):
Please check it has been updated

PR #966, opened 2023-10-26
Import Set API - Attachment Handler

aatrey882:
Another way of handling attachments when integrating with 3 party tool...


kmohit2809:
Hi @aatrey882, Thanks for your contribution. I think there is already a folder for attachments in the repository, can you pls move your files to that folder so that it is aligned in one place?

aatrey882:
@kmohit2809 it is different from normal attachment, I have modified the name of the folder... let me know if it works

kmohit2809 (changes requested):
I think you can create a folder with name 'Import Set API' and you can create aa sub folder with name as 'Attachment Handler' and place your readme.md and .js files there. In that way it will be more clear about the usage of  the scripts and it will allow others to commit their related files in the Import set API folder. 
I could only see the name change of PR but the folder name and path is still same "Attachment Handler/Read Me.md". I forgot to mention the name of readme file should be 'readme.md' as per contributing guidelines, so pls change that as well.

Post changes your folder structure should be like this:
Import Set API/Attachment Handler/readme.md
Import Set API/Attachment Handler/attachmentParser.js
Import Set API/Attachment Handler/attachmentParserUtil.js

aatrey882:
@kmohit2809 done it...

kmohit2809 (commented):
It looks good now. Thanks for being patient Ashish.

kmohit2809 (approved):

PR #974, opened 2023-10-26
Bulk update catalog item images

captainbraddles:
Update to #956. Bulk catalog item image update script. This script allows for the bulk update of catalog item pictures and icons via a fix or background script.


kmohit2809 (changes requested):
@captainbraddles thanks for your contribution. As per your script I can see that you are updating the same image/icon on multiple catalog items. Could you please update the readme file accordingly.

captainbraddles (commented):
Updated readme.md with some further details.

captainbraddles:
Hi @kmohit2809. I've updated the readme as requested to clarify that the sys_id's used can be the same or different and that the images will be updated for all catalog items that are returned in the query.

Cheers,

Brad

kmohit2809 (approved):
Looks good. Thanks for the modifications.

PR #981, opened 2023-10-27
Create Own SYSID

azeezgaa:
Case

When creating new records, ServiceNow automatically generates a unique sys_id. Though at some rare places in the system you'll notice there isn't actually a 32-character sys_id used but just a string. What the?!

So could we do this ourselves? For example on a User record, with your own firstname.lastname or your nickname? Yes, you can!

How to

You only have to apply setNewGuidValue. Note this will only work for new records, not for updating records.
I also applied autoSysFields. Just because I wanted to apply a custom Created date, so this would be my actual date of birth


SapphicFire:
Thank you for your contribution. However, this submission is not relevant to this repository. We encourage contributors to review the project's objectives and guidelines (see the README.md and CONTRIBUTING.md files) before submitting pull requests. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

This commit is missing a `readme.md`

PR #983, opened 2023-10-28
State edit for grpmem

TVani5:


SapphicFire:
Thank you for your contribution. However, it seems that the file changes in this Pull Request are incorrect or invalid. Please review the files you have changed/added and ensure that all modifications are correct, valid, and align with the project’s goals and guidelines. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one. The code snippets described do the opposite of what is described in the `readme.md` - the state will be read only when the user is in the assignment group. Also, you're not passing the string value of the assignment group, you're passing a GlideElement

TVani5 (commented):
//Create a Display Business Rule with the following code and then the following code on "On-Load" client script

//Display Business Rule Code:
/*****
g_scratchpad.grpmember = gs.getUser().isMemberOf(current.assignment_group); //This returns true or false . If user is part of the group it returns true, if user is not part of the group it returns false and assign it to scratchpad variable.
*****/

if(g_scratchpad.grpmember == false)
{
  g_form.setReadonly("state", true);
}

TVani5 (commented):
This code will make state field editable only for group members, with the help of scratchpad variable from display business rule.

PR #977, opened 2023-10-27
Use case of addOption() and removeOption() in onChange client script for priority field

chetnadev:
if priority is critical, impact can be high and medium i.e, low will be removed from choice list using removeOption()
and for other priority , impact can be high, medium and low i.e, low option will be added, using addOption()

PR #986, opened 2023-10-29
Use case of addOption() and removeOption() in onChange client script for priority field

chetnadev:
I am creating a new pull request in reference to #977  that required some changes requested by @SapphicFire.
if priority is critical, impact can be high and medium i.e, low will be removed from choice list using removeOption()
and for other priority , impact can be high, medium and low i.e, low option will be added, using addOption()


SapphicFire (approved):
Thank you for your contribution and taking on the additional feedback.

PR #991, opened 2023-10-29
A phack fest pr0423

bird-03:
This fix script is created to make a post call using scripted REST API when header type is 'application/x-www-form-urlencoded' to get the token and authenticate


Lacah:
🥳

bird-03:
Thank you so much @Laszlo ❤️ .

Feels like you are pushing to be successful with HacktoberFest2023 and I am
so Happy 🙂🤩

*Submitted one more PR in the same repo in a different folder *

   - Ap05 scheduled jobs hack2023 #992

*Abhishek Pandey*
*+91 7869881249 || ***@***.*** ***@***.***>*

PR #1005, opened 2023-10-30
Delete all records

shivamTodwal:
During application development and testing, it is common to generate numerous records in a table to evaluate various functionalities, such as workflow processes and business rules. To facilitate the removal of these test records, a background script has been developed, enabling the bulk deletion of all records within a specified table.


Lacah:
Thank you for your submission. Unfortunately we already have sever code snippets using deleteMultiple() in Fix scripts, Script Includes, GlideQuery etc. and this contribution is not adding anything more to those, and as such, I have no choice but to close it as it does not meet quality criteria for contributions.

PR #1011, opened 2023-10-30
Pull MRVS Value

azeezgaa:


Lacah:
Thank you for your contribution. However, the title and description of this Pull Request are too vague and do not provide sufficient context or clarity regarding the changes made. Please update the title and description with more detailed and specific information about the modifications, the reason behind them, and their expected impact. This will help in better understanding and reviewing your contribution effectively.
Because it's the last day of Hacktoberfest, I will not close this now (yet), but please make the updates directly in this PR. Thanks.

azeezgaa (commented):
Use this onLoad catalog client script to fetch the MRVS variable value and hide the variable.  In this example, a catalog item for blocking multiple IP addresses on a firewall has a variable **address_type** with two choices - IPV4 and IPV6. The MRVS has two variables (ipv4_address and ipv6_address) for the respective address types. If the Address type field on the parent form is set to IPV4, the field IPV6 address is hidden on the MRVS.

It returns the value of the specified field on the catalog item form when used in a client script on multi-row variable sets (MRVS).

azeezgaa:
I dont know which title and description you are referring and not sure how to change, i am not so familiar with  this tool :(

PR #1015, opened 2023-10-30
Autopopulate location

azeezgaa:
## Autopopulate Location Catalog Client Script

Use this onChange catalog client script to populate a location variable in a catalog item based on a modifiable requested_for. Both variables must be reference type pointing to their respective tables.

## Use case

The GlideUser API (g_user) does not provide a way to retrieve a user's department in a client-side script. Here is a small snippet using `g_form.getReference()` to retrieve the user from the **requested_for** variable and populate the **location** variable in a catalog item while allowing the submitter to still change the department if the data is incorrect.

Attach this client script to a variable set for easy reuse. It can be augmented with a number of other fields from the user record such as email, phone number, manager, etc. Just be mindful of field types and whether the desired field will return a sys_id or display text.


SapphicFire:
Thanks for your submission. It appears that this pull request contains more files than described in the title. Please review and update the pull request to either remove the extra/unintended files or to update the description to be more comprehensive. Ensure your contributions are meaningful and adhere to the project’s standards. You can create a new branch on your forked repository to avoid accidental inclusions of new commits in the future. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

Please note that the MRVS PR encounters the same issue as these are on the same branch. Additionally, that PR is directly copied from the documentation, please ensure you extend or differentiate those snippets to adhere to the contributing guidelines

PR #1017, opened 2023-10-30
Created : "Unpublish public reports" Scheulded Job Script

JonSandoval:
Created folder directory, readme.md, and script file. Goal of scheduled job is to look for any public reports and un-publish them from public to prevent any unintended sharing.


SapphicFire (approved):
Thanks for your contribution!

PR #1027, opened 2023-10-31
Generating a direct login URL for a user using Digest token authentication.

deepak-64742:
The script include provides functions for **generating direct login URLs for users**. This functionality is similar to the **impersonate user** feature in ServiceNow, but it allows you to directly login the user without requiring them to enter any credentials using **Digest token authentication** method.

Also I have added **overview, security factor's, installation and usage** in **readme.md** file.


SapphicFire (approved):
Thanks for your contribution!

PR #1028, opened 2023-10-31
Cancel incident

riya-misra-1:
I have created a ui action run on incident table which is client callable, run on condition current.state == '1' which means when state is new.It cancel the incident through the form.


SapphicFire (approved):
Thanks for your contribution!

PR #1030, opened 2023-10-31
Nse

Thrizvi:


SapphicFire:
As per multiple previous PRs, this contains unintended files, your description is unclear, most are missing readme.md as per contributing guidelines, and many files are in the wrong folders. Please rectify this for any future PR or otherwise these will be marked as spam.

PR #1040, opened 2023-10-31
List of managers

anubhav-ritolia:


earlduque:
not sure why you would need glideaggregate for this instead of just using gliderecord, but in the spirit of getting hacktoberfest done, i'll allow it for now.

PR #1044, opened 2023-10-31
ifConflictStopChangeRequestToAssessState

lakshmikanthgupta:
This code helps in restricting the records to move further especially for change requests if there are some conflicts in the records it also shows an error message to the user.


earlduque (changes requested):
Thank you for your contribution. However, it doesn’t comply with our contributing guidelines. As a reminder, the general requirements (as outlined in the [CONTRIBUTING.md file](https://github.com/ServiceNowDevProgram/code-snippets/blob/main/CONTRIBUTING.md)) are the following: follow the folder+subfolder guidelines

We'll keep the PR open for now in case you can make fixes, so that it can potentially still qualify for Hacktoberfest (it just had to be opened during October, it can still be approved after October). We will likely close stale PRs in one week.

PR #1047, opened 2023-10-31
Glide record branch

lakshmikanthgupta:
This code helps in gliding the required table and querying the table records based on the given conditions and if the conditions are met, will update the record as required. This helps in bulk update of the records data at a time.


Lacah:
Thank you for your contribution, though I'm afraid this submission does not meet the quality requirements of this project. It is not showcasing any particular code that is not covered by other scrips already, plus for this use case, the .updateMultiple() of GlideRecord is a better one anyway.

PR #1059, opened 2023-11-01
List of child incident

anubhav-ritolia:
This code will display all the Child Incident Numbers for the particular Incident. Updated code as per the comment.


GlideRecord/List of Child Incidents/listofchildIncident.js
```diff
@@ -0,0 +1,10 @@
+var inc = new GlideRecord('incident');
+inc.addEncodedQuery("parent_incident=9c573169c611228700193229fff72400"); // sys_id of parent Incident
```
Lacah, line 2: Sys Id should not be hardcoded in the query condition, please create a variable for it instead e.g.:
```
var parentSysId = ''; // add the sys_id of the parent incident here
...
inc.addEncodedQuery('parent_incident=' + parentSysId);
```

Lacah: Please also be consistent in using quotes vs double quotes in your code. Choose which one you want to use and stick to that for all string value definitions.

Lacah (changes requested):

anubhav-ritolia (commented):
This code will display all the Child Incident Numbers for the particular Incident.

Updated code as mentioned.

anubhav-ritolia:
This code will display all the Child Incident Numbers for the particular Incident. Updated code as per the comment.

GlideRecord/List of Child Incidents/listofchildIncident.js
```diff
@@ -0,0 +1,11 @@
+var parentID = "9c573169c611228700193229fff72400"; // This should be stored in sys_property instead of hardcoding the sys_id. 
+var inc = new GlideRecord('incident');
+inc.addEncodedQuery('parent_incident='+parentID); // sys_id of parent Incident
+inc.query();
+
+gs.info('List of Child Incidents of INC0000001:');
```
Lacah, line 6: This line makes no sense with this INC number. It should print the parent number dynamically. Something like this perhaps (added after the query statement):
```
if(inc.next()) {
  gs.info('List of child incidents of ' + inc.getDisplayValue('parent_incident'));
}
...
```

Lacah:
> This code will display all the Child Incident Numbers for the particular Incident. Updated code as per the comment.

This is just a comment not a description. You should add a description every time you open a pull request, as per the contributing guidelines. I have added this for now.

Lacah (changes requested):

anubhav-ritolia (commented):
Updated double Quotes with Single quotes and changed INC number to dynamic

GlideRecord/List of Child Incidents/listofchildIncident.js
```diff
@@ -0,0 +1,11 @@
+var parentID = '9c573169c611228700193229fff72400'; // This should be stored in sys_property instead of hardcoding the sys_id. 
+var inc = new GlideRecord('incident');
+inc.addEncodedQuery('parent_incident='+parentID); // sys_id of parent Incident
+inc.query();
+
+gs.info('List of Child Incidents of:'+inc.getDisplayValue('parent_incident'));
```
Lacah, line 6: You should have the if condition to validate the query, as I posted in my previous comment:
```
if(inc.next()) {
  gs.info('List of child incidents of ' + inc.getDisplayValue('parent_incident'));
}
```

Lacah (changes requested):

anubhav-ritolia (commented):
Added If condition as mentioned.

GlideRecord/List of Child Incidents/listofchildIncident.js
```diff
@@ -0,0 +1,14 @@
+var parentID = '9c573169c611228700193229fff72400'; // This should be stored in sys_property instead of hardcoding the sys_id. 
+var inc = new GlideRecord('incident');
+inc.addEncodedQuery('parent_incident='+parentID); // sys_id of parent Incident
+inc.query();
+
+if(inc.next()){
+gs.info('List of Child Incidents of:'+inc.getDisplayValue('parent_incident'));
+gs.info(inc.getValue('number'));
```
Lacah, line 8: What is this line here?

Lacah: Should be removed right?

Lacah (changes requested):

GlideRecord/List of Child Incidents/listofchildIncident.js
anubhav-ritolia: As we are using next(), it is already iterating a record so have to use it to display. Removing it will display 1 record less. I have tried and after removing instead of 4 child incident, it will show 3 only. So I believe we have to keep it.

Lacah: Oh yes you're right. Might be more elegant adding all the numbers to an array and printing it then, but nevermind.

Lacah (approved):

PR #1060, opened 2023-11-01
List of child incidents of all Parent Incidents

anubhav-ritolia:


Lacah:
Thank you for your submission. Unfortunately this code snippet looks quite dangerous to be used in most customer instances of ServiceNow, as it literally queries the entire incident table and then for each record found, could creates another query. This will easily cause serious performance issues that can bring down an entire node and as such it should not be added to this repository.

If you would like to share a code snippet for this use case, please optimise your code as well as extend the readme file to provide more support and context for users, for example:
- add .setLimit to the initial query to limit what is being checked (or add some query). Keep in mind that some instances can have millions of records in the incident table.
- Don't do a second GlideRecord operation embedded in the first one in a while loop. Instead you could use [GlideAggregate](https://developer.servicenow.com/dev.do#!/reference/api/utah/server/no-namespace/c_GlideAggregateScopedAPI) for this use case in one go:
  1. Run a GlideAggregate query for incidents where the parent field is not empty (again, setLimit and/or additional query conditions might make sense).
  3. Add a 'COUNT' aggregate for the parent field
  4. Order by the parent field
  5. Finally get the aggregate and print it.

This is all documented in the above link but there are also similar examples in this repository.

Please also not that in this case the code snippet should be moved to the GlideAggregate folder. Remember to add a description as well.


Normally I would close this PR, but given that Hacktoberfest is over now, I will leave it open for a while and give you the opportunity to update it. If you cannot or don't want to update it in the next days, I will close it though. Thanks.

anubhav-ritolia (commented):
Updated both files as commented.

Lacah:
> Updated both files as commented.

This is still a nested GlideRecord query and not GlideAggregate. Please review my original comment and the linked documentation carefully. This code basically needs to be rewritten as it cannot be accepted as is. Thanks.

anubhav-ritolia (commented):
Updated the code to use GlideAggregate instead of GlideRecord as suggested.

anubhav-ritolia:
Created new folder in GlideAggregate Folder.

anubhav-ritolia (commented):
Created Folder in GlideAggregate Folder.

Lacah (approved):

anubhav-ritolia:
Hi @Lacah 

My 4 Pull Request are approved but I did not got any email with confirmation that I have completed Hacktoberfest 2023 contest. Any update on this?

Lacah:
You can check the status on your hacktoberfest profile: https://hacktoberfest.com/profile/

There is a 7 day waiting period after a PR is accepted. If they were accepted, no need to worry, just have to wait.

PR #1065, opened 2024-05-10
Location hierarchy

maialithar:
Location hierarchy widget displays a location data in a hierarchical way. It is meant to be used with the catalg item on the service portal. 
It allows end user to add new locations (without saving the data in the platform) in a friendly interface. It can be adjusted to any location hierarchy.
It uses dropdown components to allow selection and modals for the addition of the data.


kmxo:
Looks good!

PR #1072, opened 2024-10-01
Remove HTML Tags from a String in a Flow

vermaamit1664:
This code will make use of a regex and JavaScript replace function to get rid of HTML tags in a string by accepting HTML string variable as an input and giving plaintext string as an output without using the Replace String transformation function of flow designer which doesn't goes well with the HTML variables.


kmxo:
Looks good to me!

szentpali-janos (commented):
I wonder if you could enhance the solution and replace the RegExp base HTML stripping with a one based on a built in function/method: `new GlideSPScriptable().stripHTML('<html>')`?
If we take for example HTML `<p>One</p><p>Two</p>`, running `.replace(/<[^>]+>/g,'')` over it produces output `OneTwo` but running `new GlideSPScriptable().stripHTML()` over it produces output `One Two`. I would say a better stripping, honoring the intent of the original HTML.

PR #1074, opened 2024-10-01
Item based user criteria

Lacah:
A script include that checks if a specific user is entitled to access a particular catalog item or not, based on the 'Available for' and 'Not available for' relations of that item.


kmxo:
Sounds good to me!

PR #1078, opened 2024-10-01
MRVS Reference Qualifier Includes Value From Catalog Item Variable

bkb68:
On a reference variable in a multi-row variable set, sometimes you want the reference qualifier to include the value of a variable that is part of the Catalog Item, not within the MRVS


snamjosh (approved):
Thanks for your contribution.

PR #1080, opened 2024-10-01
GraphQL INC Details

at8807602:
Example of groups query for GraphQL API
The code within this filder is just an example of working and basic implementation of sample group schema using GraphQL API in ServiceNow. It consist of:

1. schema (schema.gql)
2. scripted resolvers (.js extension)
3. json file representing the resolver mappings (key is Path, value is Resolver)

This is just an example
So read it, learn it and explore GraphQL 😊


at8807602:
GraphQL Basic code snippet to fetch the incident details.

snamjosh (approved):

PR #1084, opened 2024-10-01
Fix script update variable

bkb68:
a Fix Script to update a variable value on one or more RITM records


MartinStoyanoff (approved):
Looks good!

PR #1085, opened 2024-10-01
Validate a data field

kmxo:
Validate a data field in a Catalog Item via Catalog Client Script.


MartinStoyanoff (approved):
Looks good to me

PR #1088, opened 2024-10-01
Hide related list

kmxo:
Hide related list. If you have an Approval list and would like to hide it for the user who created the record then this script can help.


snamjosh:
Thank You @kmxo for your contribution. But, I see few issues in the implementation. What if logged in user tries to navigate to sysapproval_approver.LIST and approve their own record? This requirement should be handled by configuring ACLs so that it works from any UI. Please come up with a better solution to implement this requirement.

PR #1090, opened 2024-10-02
Bulk Update Function - Use the code/function to bulk change some fiel…

templetontsai:
Bulk Update Function - Use the code/function to bulk change some fields in any tables in your background script and in any other serverside scripts.


kmxo:
Sounds good to me!

PR #1091, opened 2024-10-02
Background scripts

templetontsai:
Bulk Delete Function - Use the code/function to bulk-deletes records from multiple tables based on provided encoded queries.


kmxo:
Thanks for your submission. It appears that this pull request contains more files than described in the title. 
I'll get rid of them for now but please remove the extra/unintended files within your next pull request.  
You can create a new branch on your forked repository to avoid accidental inclusions of new commits in the future.

templetontsai:
Thank you for the feedback. I will make sure the next push is from the new branch.

PR #1097, opened 2024-10-02
Br attachment variable

bkb68:
A Business Rule to make an attachment that was added to an attachment type variable appear at the top of the target record with the paper clip icon, instead of in the activity stream.


Lacah (approved):

Lacah:
Thanks for your contribution.

PR #1104, opened 2024-10-02
Form field count

bkb68:
Generate a list of forms that have more than a certain number of fields across all form sections.  These appear on the Health Scan / Impact report as a possible performance impact, with the observable behaviour being the form takes noticeably too long to load.


abhrajyotikanrar (approved):
looks good!

PR #1112, opened 2024-10-03
Utility: Validate the email address in the catalog whenever user is going to add another email

dadhich-ashish:
This is a client script that validates the email address in the MRVS field. It uses MutationObserver to observe the changes in the MRVS field.

UseCase: Validate the email address in the catalog whenever user is going to add another email, and show an error message if a duplicate email address is found.


Lacah (approved):

Lacah:
Very interesting, thanks!

PR #1114, opened 2024-10-03
Sn ap2024 01

bird-03:
Created a fix script to execute on the Log[syslog] table. This fix script will help to identify the Top 'N' number of contributors (in this case Top -10) by SOURCE name to further investigate and take action to maintain the Log table health by monitoring the pattern.


abhrajyotikanrar (approved):
Looks good!

PR #1116, opened 2024-10-03
Create Hide Attachment icon.js

riya-misra-1:
This is the catalog client script which will help us to hide the attachment icon on catalog item


abhrajyotikanrar (approved):
Looks good!

PR #1117, opened 2024-10-03
Send notification to the assigned user

shridk:
The UI Action contains a server-side script that creates a record in the sys_email table, including the email body, subject, and recipient details.
When the button on the incident form is clicked, an email will be sent to the user to whom the incident is assigned.

If the incident is not assigned to any user, a message will be shown, and no email will be sent.
If the assigned user's email address is missing, the email will not be sent, and an appropriate message will be displayed.
However, if the incident is assigned to a user with a valid email address, the UI Action will successfully send the email to the assigned user.


abhrajyotikanrar:
Hi @shridk, thank you for your contribution. However, we believe this is not a correct approach and there are better approaches available to trigger email to any user. Please follow the best practices and submit the solution again. This PR will be marked closed as of now. Feel free to create new PR while you make the necessary changes.

shridk:
Can you please explain what is not right or which best practices are not
followed?

Shrikant Kulkarni
Aaditya Software Solutions

PR #1119, opened 2024-10-03
Validate data before insert

ashoo-jindal:
Provides a utility method to perform validation checks on GlideRecord objects.


kmxo:
Sounds good to me!

PR #1120, opened 2024-10-03
Restrict user to impersonate

at8807602:
This script include helps to restrict the user from impersonation.


ashu759 (commented):
Can you please elaborate more on this script. Like the description with when this can be used and how to call this. Example it will be beneficial if you also add small set of client script for users.

Lacah:
Please update the readme file per Ashutosh's suggestions and reopen this PR, or submit as a new one. Thanks!

PR #1123, opened 2024-10-03
Fix script to regenerate the draft document if the dynamic field values used in the template have changed

aswamy93:
Leverage HR Document utils provided OOB and call these utilities to regenerate the documents. After this script runs, a new record will be created in draft_document table. The previous version of the document in draft_document will be set to inactive


aswamy93:
Commit to main branch

PR #1127, opened 2024-10-03
Dynamically update reference qualifier

bkb68:
When we have a reference variable that is used in a (single row) variable set, sometimes we want to update the Reference qualifier only for specific Catalog Item(s).  This solution replaces the common workaround of creating a Catalog Item variable with the new qualifier and hiding the variable set variable.

Now we can dynamically change a reference qualifier!!!


abhrajyotikanrar (approved):
Looks good!

PR #1132, opened 2024-10-03
Enforce Single Attachment Rule for HR Core Tasks

chetnadev:
This code enforces a rule in ServiceNow that allows only one attachment per HR core task. It checks for existing attachments, displays an error if there’s more than one, and aborts the action, ensuring data integrity and preventing confusion.


kmxo:
Sounds good to me!

PR #1137, opened 2024-10-04
Update-set cross scope checker.

RahmanM:
This sample script, checks all the WIP update-sets and find the update-sets that has more than one Application Scope in them. It will show which update-set and whose update is causing it.


RahmanM:
This sample script, checks all the WIP update-sets and find the update-sets that has more than one Application Scope in them. It will show which update-set and whose update is causing it.

PR #1139, opened 2024-10-04
Submit button

riya-misra-1:
This is the catalog client script that update the inner text of Order now button to Submit button on any catalog item


ashu759 (commented):
Can you change the tile to reflect something related to code. Also can you explain in description why this is needed. 

Finally, I will have to check internally because DOM manipulation have adverse effect on the platform and can cause issue;s.

PR #1143, opened 2024-10-04
Major incident proposal

ashoo-jindal:
Major Incident Proposal - Client Script & Script Include
This solution asks the users to propose an incident as a Major Incident candidate from the Incident form when the priority is set to 1 (Critical) and Creates a Major Incident if the user answers Yes.


earlduque:
Thanks for the detailed documentation readme!

PR #1144, opened 2024-10-04
Update Locked Out field when Active field changes on User record

aswamy93:
As per OOB, the "Locked Out" field will be set to true if Active field set to false. But, the vice-versa case is not implemented. For example, If Employee is on long leave like maternity leave etc., then the user account will be set inactive temporarily and reactivated upon his/arrival to office. This causes, user account to be Active and Locked out as true which makes the login to fail


earlduque:
Well, this isn't much of a code snippet, I'm going to approve this Pull request because I like the use case and think it's relevant for this repository. Thank you for your contribution!

PR #1161, opened 2024-10-04
Reject the ritm via fix script

at8807602:
This fix script helps to reject the RITM manually without impacting any other OOTB functionality. If in case where work flow not get triggered properly.


abhrajyotikanrar (approved):
Hi @at8807602, thank you for your contribution. We found a typo in your `Reject RITM.js` file in line nos. 6, 7, 8, 9, 14. `Approval` has been changed to `approval` as defined in line 5.

Looks good to be approved!

PR #1162, opened 2024-10-04
Send notification to the assigned user

shridk:
Astrid Sapphire kindly take a look.. Readme updated as per suggestion.


kmxo:
Sounds good to me!

PR #1165, opened 2024-10-04
Age of incident info

SrijanPatwa:
This script is designed to calculate the age of incidents in the ServiceNow platform by comparing the creation date of each incident with the current date. It retrieves all the records from the incident table and iterates through them. For each incident, it calculates the time difference between its creation date (sys_created_on) and the current date in milliseconds, then converts that value into days.


kmxo:
Sounds good to me!

PR #1166, opened 2024-10-04
Inactive user cleanup

ankitshu:
This utility is designed for ServiceNow administrators to help maintain user accounts by automatically deactivating users who have not logged in for a specified number of days. It's a Script Include that can be scheduled to run at regular intervals or executed manually.


kmxo:
Sounds good to me!

PR #1169, opened 2024-10-04
Dynamically update reference qualifier

bkb68:
Updated readme file with a key instruction that was missed earlier


earlduque:
Hey there, thank you for the contribution. I would love to include this, but I think for us to approve it We would love to see you Explain why not following this step would cause problems. For the sake of hacktoberfest we are trying to make sure changes are substantial enough to ensure people aren't trying to make spam

bkb68:
Sorry this wasn't clear in the readme.  This contribution is a solution to
a long-standing problem:  On a variable within a single-row variable set,
being unable to change a reference qualifier for a specific Catalog Item
that includes the variable set.


Without this solution, common workarounds are to either not use the
variable set, necessitating the re-creation of all of the variables, or to
create one variable with the new reference qualifier in the Catalog Item,
then hide the variable that is in the variable set for this Catalog Item.



The downside to the first workaround is re-creating potentially many
variables, while the downside to the second workaround is that the new
Catalog Item variable may appear in a different spot / order depending on
the size and construct of the variable set.  The biggest downside common to
both workarounds is that to maintain reporting on the variable(s) across
Catalog Items, and downstream usage, you would need to copy the value(s)
from the Catalog Item variable(s) to the one(s) in the variable set.  Then
keep them in sync, or maybe hide the new one(s) instead of the old one(s)
after population,...

Lacah:
Since this PR only contains a minor change in an existing readme file, I will accept and merge is, but label as invalid as it should not be counted in Hacktoberfest due to how low effort it is. Hope that makes sense to you.
Thanks.

bkb68:
Obviously the update to the readme file was low effort. Is there a
different way I should submit/ make changes to my previous submissions?

Lacah:
This is fine Brad. I only made it invalid to exclude from the Hacktoberfest counting and I am not aware of any other way (besides spam of course).
Once your PR is closed, I believe this is the best you can do, and is entirely normal (and needs no explanation outside of Hacktoberfest).

bkb68:
Thanks. I didn’t realize updating my own submission would be seen as a new
effort and that was the subject of the original question, not pertaining to
the original submission. I might get this all figured out by the end of the
month.

PR #1172, opened 2024-10-05
Enforce File Upload Restrictions for HR Document Submission

chetnadev:
This code verifies that any document uploaded for a specific HR task adheres to defined criteria: it must be in JPG or JPEG format and cannot exceed 2 MB in size. If either requirement is not met, the upload is stopped, and a relevant error message is presented to the user, ensuring data integrity during the process


Lacah (approved):

Lacah:
LGTM

PR #1178, opened 2024-10-05
Finding groups with Inactive Managers

koratalas:
This script is designed to identify all active user groups in ServiceNow where: -->The group has no manager assigned, OR
-->The assigned manager is inactive.
and store the names of the found groups in an array "arr" which can utilized later.


kmxo:

The variable called **gr** can collide with global variables with the same name. In this case it was used inside a function with is a good thing because it became a local var. 
Please try to use more descritive var names. IE: **grUserGroup**

This contribution sounds good to me!

kmxo:
Hi @koratalas,

Yes, folder created and files moved.

PR #1184, opened 2024-10-06
a function to replace Special Characters 

dcord1x:
The replaceSpecialCharacters function is designed to replace HTML special character entities in a string with their corresponding unescaped characters.

### DIFF
--- a/Utility/replaceSpecialCharacters/readme.md
+++ b/Utility/replaceSpecialCharacters/readme.md
@@ -1,0 +1,24 @@
+
+Overview
+The replaceSpecialCharacters function is designed to replace HTML special character entities in a string with their corresponding unescaped characters. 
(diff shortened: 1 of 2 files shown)


Lacah:
Thank you for your submission, but I won't be able to accept it due to 2 reasons:

1. It attempts to replicate the out of the box [unEscapeHTML method of the GlideStringUtil class](https://developer.servicenow.com/dev.do#!/reference/api/xanadu/server/no-namespace/GlideStringUtilScopedAPI#GSUS-unescapeHTML_S)
2. The Utility folder is not valid - it was accepted incorrectly in a previous PR, and will be removed. As such, it should be moved under one of the existing category folder.

If you believe that your utility has additional functionality beyond GlideStringUtil, please feel free to reopen this PR or submit a new one with additional explanation. For now, I will close this due to the above reasons.

PR #1191, opened 2024-10-06
Check for active tickets before deactivating user

koratalas:
This Business Rule checks for active tickets across multiple task-based tables (e.g., incident, problem, change request, etc.) assigned to the user before allowing the user to be deactivated. If any active tickets are found, it prevents the inactivation and displays an error message to the user.


ashu759 (approved):

ashu759:
Approved. Looks good.

PR #1201, opened 2024-10-07
Property json validator

umangkedia92:
This BR will allow to validate against the proper format for a use case where users save JSON objects in Properties. This BR will ensure that the JSON object saved is syntactically correct, and will throw an error message if it is unable to parse the JSON before saving. 

The Regex in the BR checks for most common JSON objects, and validates the structure. 

This BR will save a ton of time to debug codes where JSON objects are being used by referencing to a property.


abhrajyotikanrar (approved):
Looks good!

PR #1210, opened 2024-10-08
UI action to export incident data to csv

VaishnaviLathkar98:


kmxo:
Thank you for your contribution. However, this is an OOTB feature. 
In an Incident list, mouse over a column name and you’ll see 3 dots. 
Click on the 3 dots and you can navigate to: Export, CSV.

PR #1212, opened 2024-10-08
Added lineManagers function in CustomUserUtils Script Include

anveshmupparaju:
Added a method to the CustomUserUtils Script Include to get the Managers (Users who have at least on direct reportee).


ashu759 (approved):
Looks Good.

PR #1213, opened 2024-10-08
Shamma Negi - Calculate Aging of the Cases via Scheduled Job script

spnegi:


spnegi:
This script helps in calculating the aging of the cases and divide them in 0-30, 31-60days bucket

PR #1215, opened 2024-10-09
RM: Excel reader by row added

RahmanM:
Adding excel reader that enables processing of data row-by-row.

### DIFF
--- a/Script Includes/Excel Reader by row/ExcelReaderHelper.js
+++ b/Script Includes/Excel Reader by row/ExcelReaderHelper.js
@@ -1,0 +1,50 @@
+var ExcelReaderHelper = Class.create();
+ExcelReaderHelper.prototype = {
+
+    /**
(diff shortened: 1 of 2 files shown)


RahmanM:
Adding excel reader that enables processing of data row-by-row.

Lacah:
Thank you for your contribution. Unfortunately, I won't be able to accept it, because it is a duplicate of the already existing [Excel Parser snippet](https://github.com/ServiceNowDevProgram/code-snippets/tree/main/Script%20Includes/Excel%20Parser).
In case you think this adds some extra functionality / value, please elaborate and reopen this PR (or submit a new one).

RahmanM:
Hello,

This is different from the one already published. This enables it to
process the rows one by one rather than reading all 10000 rows and then
passing as an array...


Regards,


Rahman

PR #1216, opened 2024-10-09
RM: Request approval helper added

RahmanM:
Checks all RTIMs approval status for a request returns True if all RTIMs are approved or rejected


RahmanM:
Checks all RTIMs approval status for a request returns True if all RTIMs are approved or rejected

Lacah (approved):

Lacah:
Look good, thanks.

PR #1218, opened 2024-10-09
RM: Password generator added: Generates a random password with a specified length

RahmanM:
Generates a random password with a specified length


RahmanM:
Generates a random password with a specified length

kmxo:
There are some OOB options as we can see here:
```
var username = 'itil';
var password = SNC.PasswordPolicyEvaluator.generateUserPassword(username);
gs.info(password); // to view the password
```
But I like your code since we can specify the length.

PR #1234, opened 2024-10-09
Function to Validate IP Address both IPv4 IPv6

naveenraw25:
The `checkIPAddress` function validates whether a given **string is a valid IPv4 or IPv6 or Invalid address**. It also includes a function to check if the string is a valid hexadecimal number.

- Created functions to validate IPv4 and IPv6 addresses.
- Added comments to explain each part of the code to enhance readability and understanding.
- Included examples to demonstrate how to use the validation function and logging for validation results.
- Created a README file with a title, description, usage instructions, and examples for the IP address validation function.


snamjosh:
Thank You @naveenraw25 for your contribution. But, please note that this is available OOB with regex.  You can reuse OOB IP regex pattern instead of writing custom code.

PR #1248, opened 2024-10-10
Autopopulate Caller Location in Short Description

jahnaviT2003:


kmxo:
Hi, 

First of all please let me give some advice:
1) Probably you'll need to use global.AbstractAjaxProcessor in order to use an Ajax call. 
2) When we use getXML() method we get a lot of fields within the payload. Compare it with the method getXMLAnswer() to see which is more efficient. 
3) While opening a Pull Request, always provide a good **description** about the new feature.

Your readme file is really good.

PR #1254, opened 2024-10-10
Auto numbering issues fixscript

Mexiprince:
Auto numbering issues fixscript, readme and codefile


404paige:
Hello! Thank you for your contribution. There are a few errors in the script you submitted. Additionally there are some best practices not being followed (such as using gr as a variable name). 

Lastly, I do not think running this code as currently written would be advisable. Generally in a scenario where numbering might have gone askew you'd want to either query for all the records with duplicate numbers and renumber one or both records or provide the duplicate records and have them renumbered. The script as written would renumber every record in a table. This could be incredibly confusing for end users (say, for example, you ran this on the incident table... users would know INC000123 is my incident and suddenly that may no longer be correct).

Please reconsider the submitted code. Thank you!

PR #1255, opened 2024-10-10
Get contains roles of a role

Praiseto:
The script in "Get contains role of a role.js" file will get all the roles contains in a role.

The <role_name>' needs to be replaced with the name of a role to which you need to get all the contains role


ashu759 (approved):
Looks good.

PR #1264, opened 2024-10-11
Shamma Negi - Create attachment via Script in ServiceNow

spnegi:
With this script you can create any type of attachment like text,doc,xlsx. This Agent "AttachmentCreator" help you to create the attachment vua ecc_queue. As soon as this script is called an entry is inserted in ecc_queue table with ouput type then once the file is create you will receive another entry as input with status as ready.

You can call this script from any server side scripting. It can help you create any number of attachments.


kmxo:
For a future version you can research about how to provide other file types like docx or pdf.

PR #1265, opened 2024-10-12
Adding bookmark to favorites tab

koratalas:
The script is designed to bookmark the list of incidents assigned to the user's groups in ServiceNow's favorite tab if not already present.


Background Scripts/Adding bookmark to Favorites tab/Adding bookmark into Favorites tab.js
```diff
@@ -0,0 +1,28 @@
+var memGroups = []; // empty array for getting the group sys'id rin future.
```
wiz0floyd, line 1: FYI there is an OOB Dynamic query for My Groups so it's not needed to query the group table and build this array.

PR #1266, opened 2024-10-12
fix: Get URL Parameters

### LINKED ISSUE #416: window == null to determine if in service portal not working

ronaldkarim:
Refactor - Indenting And Whitespace

Replace check for window object with check for spModal object,
which is available in Service Portal only.

Update URLSearchParam from sysparm_id to sys_id.

Resolves #416

### DIFF
--- a/Client Scripts/Get URL Parameters/script.js
+++ b/Client Scripts/Get URL Parameters/script.js
@@ -1,15 +1,14 @@
 function onLoad() {
-//Type appropriate comment here, and begin script below
+    //Type appropriate comment here, and begin script below
 
-if(window == null){ // For Service Portal
-var url = top.location.href;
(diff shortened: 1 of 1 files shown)


SapphicFire (approved):
Looks good to me.

PR #1274, opened 2024-10-13
Catalog Item Explorer Widget - Support of multiple catalogs has been added

ivan-betev:
Widget can now be used to list items from multiple catalogs for that I have added some changes in client/server sides and updated options schema.


SapphicFire (approved):
Looks good to me :)

PR #1277, opened 2024-10-13
Get encoded query in readable format

priyasnexpert:
This background script from Pull request is used to get the any encoded query in readable format.


SapphicFire (approved):
Looks good to me

PR #1278, opened 2024-10-13
Get day of week as per local time from GMT

priyasnexpert:
The code contains this Pull Request is to get the day of the week as per your local time from GMT.


SapphicFire (changes requested):
Thank you for your contribution. However, it seems that the file changes in this Pull Request are incorrect or invalid. Please review the files you have changed/added and ensure that all modifications are correct, valid, and align with the project’s goals and guidelines. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

This appears to find the offset between a provided time which is purportedly in GMT and instance time (Pacific Time out of the box), making the assumption that it will always be a negative UTC timezone, and then converting this to positive and giving a local day/time that is not the case (1AM GMT would be 6PM UTC-7 on the previous day for example).

Is this script intended to do the inverse? If taking a time locally and converting to GMT, this makes more sense.

I look forward to your update and approving this PR.

PR #1282, opened 2024-10-13
Client Script for Label change and GlideAggregate Code to find Create Active SLA Count by Assignment Group

nowitsvashu:
This commit contains two files for Client Script for Label change and GlideAggregate Code to find Create Active SLA Count by Assignment Group.


MartinStoyanoff:
@nowitsvashu there are two scripts and two readme files on that one, please split them into two separate pull requests. thanks!

PR #1297, opened 2024-10-14
Active sla count by assignment group

nowitsvashu:
This commit contains the code for Active sla count by assignment group using the glideaggregate


kmxo:
Sounds good to me.

PR #1300, opened 2024-10-14
Add the members to List of the Groups using GlideRecord

ravichandra1998g:
This is a sample Background Script to Add a User to list of required User Groups Where They Are Not Already a Member.


kmxo:
Sounds good to me.

PR #1306, opened 2024-10-14
Shamma Negi - Call Javascript Probe from Server Scripting

spnegi:
With this script you can call MID Server Script Include via JAVASCRIPT probe. As soon as you run this script it will insert the entry in
ecc_queue table and output record will be created against that. In that output record you will see the parameters sent to "SFTP"
Then once the file is successfully copied or moved to SFTP and input record in ecc_queue will be inserted and gives the output whether the file
transfer was successful or any errors.

You can call this script from any Server side scripting


SapphicFire (approved):
Looks good, thank you for your contribution

PR #1314, opened 2024-10-14
Calling pop up box in list view

YTejaswini18:
This tells how the ui action that calls a ui page from the list view of a table in ServiceNow. Specifically, it demonstrates how to open a modal dialog when multiple items are selected in the list view. The modal dialog will display a UI page and pass selected record sys_ids as parameters. Users can quickly perform actions on multiple records at once, reducing the need for repetitive manual operations.


YTejaswini18:
Provided detailed description of how the script works and the purpose of the ui action and also attached the script of ui action in Create calling_pop_up_box_in_list_view.js

PR #1320, opened 2024-10-14
Shamma Negi - Parse Excel In ServiceNow

spnegi:

This is really beneficial when bul data needs to be uploaded in system.


sandeepd26:
Thank you for your contribution. However, it seems that the file changes in this Pull Request are incorrect or invalid. Please review the files you have changed/added and ensure that all modifications are correct, valid, and align with the project’s goals and guidelines. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one. 
You have past commits reflecting in the same PR and per description the commit doesnot match to the files changed. Kindly create another PR.

PR #1321, opened 2024-10-14
Pull empty serial number asset records

YTejaswini18:
The getAssetRecord Script Include is designed to identify records in the alm_asset table that are missing a serial_number. This functionality is crucial for maintaining the integrity of asset management data, as it ensures that all assets have complete and accurate information, you can use same script include in many ways by changing the table name and the encoded query according to your requiremnt


SapphicFire (approved):
Looks good to me

PR #1328, opened 2024-10-15
Slient Update on GlideRecord

QI-XING:
Use of GlideRecordObject.autoSysFields(false) to silently update GlideRecord without system updates during background script run.


QI-XING:
Use of GlideRecordObject.autoSysFields(false)

kmxo:
Sounds good to me!

PR #1339, opened 2024-10-15
Create AssignRoletoUser

Thrizvi:
This script does the following:

Creates a new GlideRecord object for the sys_user table. Queries the table for the user with the specified userId. Checks if the user already has the specified role using the hasRole() method. If the user doesn't have the role, it adds the role to the user using the addRole() method and updates the user record. Logs a message to the console indicating whether the role was assigned or if the user already had the role.

### DIFF
--- a/Background Scripts/AssignRoletoUser
+++ b/Background Scripts/AssignRoletoUser
@@ -1,0 +1,19 @@
+function assignRoleToUser(userId, roleName) {
+    var userGR = new GlideRecord('sys_user');
+    userGR.get(userId);
+
+    // Check if the user already has the role
+    if (!userGR.hasRole(roleName)) {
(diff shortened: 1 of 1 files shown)


kmxo:
Thank you for your contribution. The idea of automate the process is good but it is not considered best practive to assign roles directly to users. 

The proper way to do it:
1 - Create a User Group 
2 - Assign the role(s) to the Group 
3 - Assign users to the Group 

The users assigned to the Group will inherit the roles. 

For the maintenance view let's say you have a Persona who have 5 roles within your system. 
You just need to assign a user to that Group and he/she will be granted all necessary roles. 
On the other hand, if the user assumes another Persona you can just remove that user from Group A and insert in Group B.

PR #1340, opened 2024-10-15
Shamma Negi - Excel Parser in MID Server Script Include

spnegi:

This is really beneficial when bul data needs to be uploaded in system.


Lacah:
Thank you for your contribution. However, it seems that the file changes in this Pull Request are incorrect or invalid. It appears that this request contains the deletion of your previous contribution, as well as additional of this parser. Is this intentionals?
Please review the files you have changed/added and ensure that all modifications are correct, valid, and align with the project’s goals and guidelines. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

spnegi:
Apologies I didnt know how the other items are getting captured in my new
branch.

I will submit a fresh one for excel parser wherein only 2 commit woll be
there

Thanks for your help!

PR #102, opened 2021-10-03
Create excelParser.js

Lacah:
Pass in the sys_id of an Excel attachment and get back an array of objects, with every row in the sheet being an element in the array.

PR #1345, opened 2024-10-15
spnegi - Parse Excel and create/update records in ServiceNow

spnegi:

This is really beneficial when bulk data needs to be uploaded in system.

### DIFF
--- a/Flow Actions/Excel Parser/Excel Parser.js
+++ b/Flow Actions/Excel Parser/Excel Parser.js
@@ -1,7 +1,5 @@
 (function execute() {
-    //
-
-    var arr = [];
+       var arr = [];
 
     var gratt = new GlideRecord('sys_attachment');
     gratt.addQuery('sys_id', "d82c84202bac5a10cac8fb806e91bf53"); //sys_id of the attachment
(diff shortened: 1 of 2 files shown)


Flow Actions/Excel Parser/Excel Parser.js
```diff
@@ -0,0 +1,38 @@
...
+        parser.setNullToEmpty(true);
+        parser.parse(attachmentStream);
+        var headers = parser.getColumnHeaders();
+        //print headers
+
+        var rows_info = [];
+        // Iterate over each row in the worksheet
+        while (parser.next()) {
+            row = parser.getRow();
+            gs.info("Shamma row" +headers.length);
```
ashu759, line 24: Clean up the code please. Like remove comments and logs file. Clean code is required.

ashu759 (commented):
Also, i read what you have mentioned in the readme file but that all does not reflect in your code. I think there a better ways to update and create records in ServiceNow like Transform maps.

spnegi:
Sure let me remove the comments and clean the code and let you know.

We can also set the parent child relationship via excel parser.

spnegi:
Hi @ashu759 , 

I have removed the comments and info log lines from the code. Also updated the readme.md.
Please verify now

Thanks,
Shamma

ashu759:
Hello @spnegi , When we run duplicate check, we already see that this functionality was already committed https://github.com/ServiceNowDevProgram/code-snippets/pull/102. So for now we are closing this pull request as we dont see much value addition in this.

PR #1364, opened 2024-10-16
Spnegi - Calculate Aging of the cases

spnegi:
This script helps is calculating the aging of the ticket/cases and define them in bucket of category aging like '0-2 Days','3-4 Days'
Based on this you can get the reporting on cases aging. How old is the case. It calculates the aging from the creation date.

It works on all cases except the cases which are on resolved,cancelled and closed state.
With this script you can decide whether to show that case in red/orange or yellow colour so that agent will know just by seeing the case
that aging has increased. So if aging is greater than 30 we can make the case highlighted as red by using field styles conditions.


sandeepd26:
Incorrect Filename !!

image

PR #1374, opened 2024-10-17
Find the reports assigned to Inactive Users

Paulsylo25:
Added a code snippet to find the reports assigned to inactive Users, This code will help you to retrieve the reports assigned to inactive users, so you can cleanup your reports


kmxo:
Sounds good to me!

PR #1383, opened 2024-10-17
Prevent duplicate update sets

bgrk005:
Adding new Business Rule to Prevent duplicate names for update sets.

1.  this will validate if there is any update set exist with name of the current update set 
2. if there is any existing update srt with same name it will display error and abort the action.


wiz0floyd:
Hi,

I've renamed your pull request. In the future please make sure to give it a descriptive name.

wiz0floyd (approved):

PR #1399, opened 2024-10-18
change state if RITM state changes

Adil404:
This business rule is to set state of sc_task table when sc_req_item state changes.
Create before insert/update business rule on sc_req_item table and give condition if state changes to 2 OR cat_item.sc.catalogs is "take your choice".

### DIFF
--- a/Business Rules/RITM state change/Related_task_state_update.js
+++ b/Business Rules/RITM state change/Related_task_state_update.js
@@ -1,0 +1,16 @@
+//create business rule on sc_req_item table and add condtion 
+
+(function executeRule(current, previous /*null when async*/ ) {
+
+    var gr = new GlideRecord("sc_task");
+    gr.addQuery('request_item', current.sys_id);
+    gr.orderBy('order');
+    gr.setLimit(1);
+    gr.query();
+    if (gr.next() && gr.getValue('state') == 1) {
+        gr.setValue('state', '2');
+        gr.update();
+    }
+
+
+})(current, previous);
--- a/Business Rules/RITM state change/Related_task_state_update.js
+++ b/Business Rules/RITM state change/Related_task_state_update.js
@@ -1,16 +1,15 @@
-//create business rule on sc_req_item table and add condtion 
-
-(function executeRule(current, previous /*null when async*/ ) {
+(function executeRule(current, previous /*null when async*/) {
+    var OPEN_STATE = 1;
+    var WORK_IN_PROGRESS_STATE = 2;
 
     var gr = new GlideRecord("sc_task");
     gr.addQuery('request_item', current.sys_id);
-    gr.orderBy('order');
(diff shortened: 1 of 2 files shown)


Business Rules/RITM state change/Related_task_state_update.js
```diff
@@ -0,0 +1,16 @@
+//create business rule on sc_req_item table and add condtion 
+
+(function executeRule(current, previous /*null when async*/ ) {
+
+    var gr = new GlideRecord("sc_task");
+    gr.addQuery('request_item', current.sys_id);
+    gr.orderBy('order');
+    gr.setLimit(1);
+    gr.query();
+    if (gr.next() && gr.getValue('state') == 1) {
```
msdfib, line 10: If you added this condition to the query, you would not need to check the state for '1' (the magic number).

Business Rules/RITM state change/Related_task_state_update.js
```diff
@@ -0,0 +1,16 @@
+//create business rule on sc_req_item table and add condtion 
+
+(function executeRule(current, previous /*null when async*/ ) {
+
+    var gr = new GlideRecord("sc_task");
+    gr.addQuery('request_item', current.sys_id);
+    gr.orderBy('order');
+    gr.setLimit(1);
```
msdfib, line 8: What if more than one sc_task is under RITM?

Business Rules/RITM state change/Related_task_state_update.js
```diff
@@ -0,0 +1,16 @@
+//create business rule on sc_req_item table and add condtion 
+
+(function executeRule(current, previous /*null when async*/ ) {
+
+    var gr = new GlideRecord("sc_task");
+    gr.addQuery('request_item', current.sys_id);
+    gr.orderBy('order');
```
msdfib, line 7: You don't need to sort the results if you don't have a clear reason, in this case, you are sorting a list with one record.

Lacah (changes requested):
Have a look at the added comments and resubmit your script accordingly. Thanks.

Adil404 (commented):
updated the script : sc_task will change the state from open to work in progress when RITM state changes to work in progress (set this condition in filer condition).

Business Rules/RITM state change/Related_task_state_update.js
Adil404: removed the orderBy query

Business Rules/RITM state change/Related_task_state_update.js
Adil404: removed setLimit(1) so that it will update all the sc_task related to RITM

Adil404 (commented):
update the script, please review it and let me know

msdfib (approved):
LGTM, leaving merge up to @Lacah

PR #1417, opened 2024-10-20
TechtrekwithAJ cmdb ci count in each class

techTrekwithAJ:
Here, this code will help to get the count of CI in each CMDB class


am-shakeel (approved):
looks good

PR #1423, opened 2024-10-20
Add signature and update fields to a fillable pdf document

gowdah:
It's a Flow Designer action that takes a record's sys_id, where the PDF record is available, and a signature as inputs, and adds the signature to the PDF.


am-shakeel (approved):
Looks good

PR #1425, opened 2024-10-20
Script Atf Step To count Table Records

gowdah:
Use this ATF script step to find number of records in a table(by passing table name in the script)


Lacah (approved):

Lacah:
Cool!

PR #1433, opened 2024-10-21
Add Users To Groups using Server-Side Scripts

prashantmrshine:


GlideRecord/Add n number of users to n number of groups using server scripts/AddUserstoGroups.js
```diff
@@ -0,0 +1,19 @@
...
+
+        var grMember = new GlideRecord('sys_user_grmember'); // Query for Group Membership table
+        grMember.initialize();
+        grMember.group = groupSysId;
+        grMember.user = userSysId;
+        grMember.insert(); // Users will be inserted into the Group Membership table
+
+        gs.info('Added User ' + userSysId + 'to group ' + groupSysId); // Info Message displaying after each users group membership creation
+    }
+
```
msdfib, line 19: Missing bracket

msdfib:
A similar script is already in repository #1300, although I like your version better.

GlideRecord/Add n number of users to n number of groups using server scripts/AddUserstoGroups.js
prashantmrshine: I have updated the bracket on my script

SapphicFire (approved):
Similar to other submissions but improves and works with multiple users and groups.

Looks good to me, thank you

PR #1438, opened 2024-10-21
emailFormatValidator can be used while importing data and conform to a standard email format.

maheshkhatal27:
The purpose of this script is to validate email addresses during a data import in ServiceNow. It ensures that the email addresses conform to a standard format (i.e., are valid email addresses), converts them to lowercase, trims any extra spaces, and logs invalid email addresses for review or further action.


SapphicFire (approved):
While this is not the most robust email regex, this is a good submission.

Looks good to me

PR #1445, opened 2024-10-21
Update field value in bulk

Adil404:
for an instance we have 100 records in incident table which state has open, now we want 100 records to be set as work in progress this fix script will help in updating field value.

### DIFF
--- a/Fix scripts/update field value/updateFieldValue.js
+++ b/Fix scripts/update field value/updateFieldValue.js
@@ -1,0 +1,8 @@
+var grTable = new GlideRecord('table_name');
+grTable.addEncodedQuery('add query according to your requirement'); // on table list filter condition right click and copy query
+grTable.query();
+gs.info('Table Records to update: ' + grTable.getRowCount());
+while (grTable.next()) {
+    grTable.field_name = '1';   // field which needs to be updated
+    grTable.update();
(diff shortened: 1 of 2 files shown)


Fix scripts/update field value/updateFieldValue.js
```diff
@@ -0,0 +1,8 @@
+var grTable = new GlideRecord('table_name');
+grTable.addEncodedQuery('add query according to your requirement'); // on table list filter condition right click and copy query
+grTable.query();
+gs.info('Table Records to update: ' + grTable.getRowCount());
+while (grTable.next()) {
```
msdfib, line 5: There is a dedicated function to bulk update - updateMultiple

msdfib:
I believe it is already in the repository -> #75

SapphicFire (changes requested):
It appears that this contribution is a duplicate or resembles spam, potentially aiming to game the Hacktoberfest event. As a reminder, submissions marked as spam may lead to account restrictions or disqualifications. Please ensure to make unique and meaningful contributions, and refrain from spamming or duplicating submissions. Closing this as spam.

Per `@msdfib`'s comments, this duplicates existing functionality in GlideRecord and does not make use of designated functionality

PR #1447, opened 2024-10-21
Getting a list of records based on user's location

prashantmrshine:


GlideRecord/Display list of records based on Users Location/listOfRecordsBasedOnLocation.js
```diff
@@ -0,0 +1,17 @@
...
+Scripts - 
+(function executeRule(current, previous /*null when async*/ ) {
+
+    // Add your code here
+    var current_user = gs.getUserID();
+
+    var usr = new GlideRecord('sys_user'); // Query User Table
+    usr.addQuery('sys_id', current_user);
+    usr.query();
+    if (usr.next()) {
```
msdfib, line 14: It is worth not to skip gr.setLimit(1) if you want to retrieve only 1 record & validate list via if(gr.next()) statement

PR #1449, opened 2024-10-21
Field Placeholder 

Srikanth1306:
Hint that appears inside an input field to indicate the type of data the user should enter. It is typically shown in a lighter font colour and disappears once the user starts typing. This helps guide users by providing an example or context without taking up additional space on the form.
For example, a field placeholder in an short description input might read: "Please give the issue details here"
image


am-shakeel (approved):
Looks good

PR #1453, opened 2024-10-22
Copy attachment from table to other table

prashantmrshine:


kmohit2809:
Hi @prashantmrshine, Thanks for your contribution. However, there is already a [code snippet](https://github.com/ServiceNowDevProgram/code-snippets/blob/main/Business%20Rules/Copy%20attachments%20from%20idea%20to%20demand/copy%20attach%20from%20idea%20to%20demand.js) which uses the same method to copy attachments. I would suggest you to verify once before adding a new snippet. In this way we can avoid duplicate snippets. Thanks for understanding. I will close this PR now.

prashantmrshine:
> Hi @prashantmrshine, Thanks for your contribution. However, there is already a [code snippet](https://github.com/ServiceNowDevProgram/code-snippets/blob/main/Business%20Rules/Copy%20attachments%20from%20idea%20to%20demand/copy%20attach%20from%20idea%20to%20demand.js) which uses the same method to copy attachments. I would suggest you to verify once before adding a new snippet. In this way we can avoid duplicate snippets. Thanks for understanding. I will close this PR now.

How do I check whether any code is already present I searched but did not find it, that's why I have added this one.

kmohit2809:
You can always do a quick search on github by using "go to files" search area present in left side of Add file option on your repository. You can use the folder name or title of file to search it. Once you hit similar kind of results, you can check if the code/functionality is same or different.

PR #1460, opened 2024-10-22
Replace Text

Jaspalsinghchot:
Purpose of the script is to help replace text IT_SAP with IT_ERP for all the Groups in Group table containing IT_SAP as keyword.
It uses replace method which follows format replace(/'ABCD'/g,'PQR')
where ABCD is to be replaced with PQR


Background Scripts/Replace Text/readme.md
kmohit2809, line 1: Hi @Jaspalsinghchot, thanks for your contribution. Could you please fix the spelling error in your readme.md file. In line number 3. And recommit it.

Jaspalsinghchot: Done

PR #1480, opened 2024-10-24
UX Data broker to fetch system property in UI Builder

TRIGGEREDcoder:
This script can be used in UX Data Broker transform script which once saved will be shown as a data resource in UIB experience page. Once it's added to the page, the user can simply provide the property name to fetch the property value from sys_properties table in the platform. which can be subsequently used in state parameters or in any component.


SapphicFire (approved):
Accepting this, but consider documenting and discussing the use cases and value of a broker like this in future

TRIGGEREDcoder:
Thanks for the feedback.

PR #1482, opened 2024-10-24
Get Duplicates

Jaspalsinghchot:
Using GlideAggregate function to find out tickets (tasks) with same number. OOB there happens to be a Unique checkbox at dictionary level and if in case not set to True it might create duplicate numbered tickets.
Script will help find, tickets if any.


kmxo:
Sounds good to me. Thank you for your contribution!

PR #1483, opened 2024-10-24
Get users based on role

Charanjet:


Background Scripts/listUsersHavingSpecificRole/script.js
```diff
@@ -0,0 +1,14 @@
+function getUsersBasedOnrole(role){
+	var userGR = new GlideRecord('sys_user');
+	var roleGR = new GlideRecord('sys_user_has_role');
+	var itilUsers = [];
+	roleGR.addQuery('role.name', role);
+	roleGR.query();
+	
+	while (roleGR.next()) {
+		userGR.get(roleGR.user);
```
wiz0floyd, line 9: Instead of making a second call to the database you could use getDisplayValue from the initial lookup of user roles.

wiz0floyd (changes requested):

Background Scripts/listUsersHavingSpecificRole/script.js
```diff
@@ -0,0 +1,11 @@
+function getUsersBasedOnrole(role){
+	var itilUsers = [];
+	var userGR = new GlideRecord('sys_user');
+	userGR.addEncodedQuery('roles.name=itil');
```
wiz0floyd, line 4: Have you tested this query? I'm not sure it will work since roles is a list field

wiz0floyd (changes requested):

Background Scripts/listUsersHavingSpecificRole/script.js
Charanjet: Yes Wiz,
I have tested this function is working completely fine and returns the list of users having itil role

PR #1488, opened 2024-10-25
Get users based on role

Charanjet:
This function will provide the list of the users with specific roles assigned to them. The function is tested and in use by our team members. We are using it on daily basis on Xanadu and Washington release.


Background Scripts/listUsersHavingSpecificRole/script.js
```diff
@@ -0,0 +1,11 @@
+function getUsersBasedOnrole(role){
+	var itilUsers = [];
+	var userGR = new GlideRecord('sys_user');
+	userGR.addEncodedQuery('roles.name=itil');
```
wiz0floyd, line 4: This should use the function parameter instead of a static value. Additionally, please confirm that dot walking like this works in a list field.

wiz0floyd (changes requested):

PR #1489, opened 2024-10-25
Approve on behalf Scripted Rest API

amigujarathi:
This REST API allows authorized users to approve or reject tasks on behalf of another user. The script handles impersonation, performs action on approval records, and returns appropriate responses based on the success or failure of the request.


am-shakeel (approved):
Looks good

PR #1500, opened 2024-10-26
Fetch requested by user for a change record using getRefRecord()

Soumyadeep10:
This script will glide for a particular record whose sys id has been provided
It will fetch the record and then use getRefRecord() to fetch the requested by user details.
If the user is a valid record, then it will fetch the user_name and email.


am-shakeel:
This PR is also seen as similar to the https://github.com/ServiceNowDevProgram/code-snippets/pull/334

am-shakeel:
This is the 3rd PR and all are seen as similar to the existing PRs. Please avoid duplication.

am-shakeel (approved):
This code will help for different use cases. Proceeding further.

PR #1502, opened 2024-10-26
Get the current version of any application.

shreyawani98:
Get the current version of any application through background script.


kmxo:
Sounds good to me!

PR #1503, opened 2024-10-26
Add bulk users to a Visual task Board

shreyawani98:
Through this fix script, we can add multiple users to a Visual Task board at once.


kmxo:
Sounds good to me. Thank you for your contribution!

PR #1505, opened 2024-10-26
Creating PDF from a Knowledge Article, using OOTB API

manrick:
There is an easier way to generate PDF from a Knowledge Article using OOTB API, called [PDFGenerationAPI](https://developer.servicenow.com/dev.do#!/reference/api/xanadu/server/sn_pdfgeneratorutils-namespace/PDFGenerationAPIBothAPI#P-PDFGenerationAPI?navFilter=pdf)

I think it's always better to use OOTB solutions, if it's available and meets the requirements.


am-shakeel:
Hi @manrick , Thanks for the PR! 
While using OOB features is a great approach, Hacktoberfest encourages original contributions

manrick:
Hi @am-shakeel, what you wrote is true and I agree with you, but in this case I have a bit different view. There is a solution on the Github, which is not bad, but there is a better one. ServiceNow always recommends to stay as close as the OOTB solution as it is possible. I have upload this code snippet, because I wanted to highlight that there is another solution for exporting KB article into PDF, which might be better or easier.
If you don't agree with me, please reject this pull request and I will submit another one.

PR #1507, opened 2024-10-26
getRecordXML Function

prithvibalakka:
This function, getRecordXML, is a custom JavaScript function designed for ServiceNow to generate an XML representation of a record in any specified table.


Background Scripts/Get XML Of a Record/getRecordXML.js
```diff
@@ -0,0 +1,22 @@
...
+    for (var i = 0; i < fields.size(); i++) {
+        fieldNames.push(fields.get(i).getName());
+    }
+    fieldNames.sort();
+
+    fieldNames.forEach(function(fieldName) {
+        var fieldValue = gr.getValue(fieldName);
+        var fieldTag = xmlDoc.createElement(fieldName);
+        fieldTag.setTextContent(fieldValue);
+        baseTag.appendChild(fieldTag);
```
wiz0floyd, line 18: This is not a documented method

wiz0floyd (changes requested):
Thank you for your submission. There are a number of things that need to be addressed. Out of curiosity, where did you find some of the methods you're using?

PR #1510, opened 2024-10-27
Passport validity

shreyawani98:
This Client Script will ensure that the entered passport information and expiry date meet the requirements, providing a seamless and guided experience for the user.

The conditions are as follows:

The passport number must be exactly 8 characters long. The first character should be an uppercase letter, followed by two numeric characters—the first between 1 and 9, and the second between 0 and 9.

The date of expiry should auto-populate based on the date of issue. If the user is 18 years or older, the expiry date should be exactly 5 years from the date of issue. If the user is under 18, the expiry date should be exactly 10 years from the date of issue.

### DIFF
--- a/Catalog Client Script/Passport Validation/passportvalidity.js
+++ b/Catalog Client Script/Passport Validation/passportvalidity.js
@@ -1,0 +1,38 @@
+function onChange(control, oldValue, newValue, isLoading) {
+    if (isLoading) return;
+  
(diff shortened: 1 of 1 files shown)


shreyawani98:
[readme.md](https://github.com/user-attachments/files/17533263/readme.md)

shreyawani98:
Hi team, I am unable to create a new description(md) file in this branch, hence attaching the file from my device.

PR #1511, opened 2024-10-27
Passport information validation

shreyawani98:
This Client Script will ensure that the entered passport information and expiry date meet the requirements, providing a seamless and guided experience for the user.

The conditions are as follows:

The passport number must be exactly 8 characters long. The first character should be an uppercase letter, followed by two numeric characters—the first between 1 and 9, and the second between 0 and 9.

The date of expiry should auto-populate based on the date of issue. If the user is 18 years or older, the expiry date should be exactly 5 years from the date of issue. If the user is under 18, the expiry date should be exactly 10 years from the date of issue.

### DIFF
--- a/Catalog Client Script/Passport Validation/passportvalidity.js
+++ b/Catalog Client Script/Passport Validation/passportvalidity.js
@@ -1,0 +1,38 @@
+function onChange(control, oldValue, newValue, isLoading) {
+    if (isLoading) return;
+  
+    var passportNumber = g_form.getValue('passport_number');
+    var dateOfIssue = g_form.getValue('date_of_issue');
+    var age = parseInt(g_form.getValue('age'), 10);
+    var dateOfExpiry = g_form.getValue('date_of_expiry');
+  
+    // Passport Number Validation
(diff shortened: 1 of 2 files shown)


am-shakeel:
@shreyawani98, As it is a onChange client script, please let me know which field we should write?

am-shakeel:
@shreyawani98  and also please provide the sequence of the fields like which field should the user fill in first.

shreyawani98:
Hi @am-shakeel 

For this onChange Client Script, we should be writing on the following fields:

Passport Number Field (passport_number): This will trigger the script to validate the format of the passport number whenever it changes.

Date of Issue Field (date_of_issue): This will trigger the script to auto-populate the date of expiry based on the date of issue.

Age Field (age): This will trigger the script to check If the user is 18 years or older, the expiry date should be exactly 5 years from the date of issue. If the user is under 18, the expiry date should be exactly 10 years from the date of issue.

The sequence of the fields should be as follows : 
1. Passport Number
2. Date of Issue
3. Age
4. Date of expiry (will be autopopulated based on the date of issue)

earlduque (approved):

PR #1512, opened 2024-10-27
Ts drawer button

stwinkle:
I have created drawer buttons for the portal homepage to enhance the user experience. These buttons provide users with additional options and quick access to links for easy redirection.


stwinkle:
Output Image:
image

am-shakeel (approved):
Looks good

PR #1514, opened 2024-10-27
Send notification if the incident remains unassigned for more than 5 days

shreyawani98:
This UI Action button will notify the assignment group's manager if the incident is unassigned for more than 5 days, helping ensure timely assignment and resolution.


am-shakeel (changes requested):
var unassignedDuration = gs.daysAgo(current.sys_created_on);

Can you please check this line and let me know, what value you are expecting in "current.sys_created_on"

PR #1515, opened 2024-10-27
Improve incident handling

shreyawani98:
Suppose you want to gather data about incident resolution in your system. Specifically, you need to find the total number of incidents, the average time to resolution (in hours), and the number of incidents per assignment group. This information can help analyze the efficiency of different groups and improve incident handling.

This GlideAggregate example provides a consolidated view of incident resolution statistics, which can aid in optimizing group efficiency and improving response times.

It fetches total counts, average resolution time, and group-specific counts.
Logs the total number of resolved incidents and the average resolution time.
For each assignment group, it logs the group’s sys_id and the corresponding incident count.


kmxo:
Since the state field is inherited from the Task table and it is an Integer, I believe it would be better to try something like
`inc.addQuery('state', 6); //state = Resolved`

shreyawani98:
Thanks for the suggestion @kmxo.I have updated the script accordingly.

Lacah:
Looks good, thanks for this contribution!

PR #1516, opened 2024-10-27
Prevent unnecessary notifications from being sent out

koratalas:
Created a background script to prevent unnecessary notifications from being sent out. It helps in managing the volume of emails being sent so that we do not send the notifications even by mistake. This script is mostly used in dev or uat to avoid any notifications being sent from lower instances.

### DIFF
--- a/Background Scripts/Prevent unnecessary notifications from being sent out/Prevent_unnecessary_notifications_from_being_sent_out.js
+++ b/Background Scripts/Prevent unnecessary notifications from being sent out/Prevent_unnecessary_notifications_from_being_sent_out.js
@@ -10,7 +10,7 @@ emailGR.query();
 while (emailGR.next()) {
     emailGR.state = "ignored"; //setting state to "ignored"
     emailGR.type = 'send-ignored'; // Set the type to 'ignored'
-    emailGR.update(); // Save the changes
+    emailGR.updateMultiple(); // Save the changes
 }
 
 gs.info('All relevant emails have been marked as ignored.');
--- a/Background Scripts/Prevent unnecessary notifications from being sent out/Prevent_unnecessary_notifications_from_being_sent_out.js
+++ b/Background Scripts/Prevent unnecessary notifications from being sent out/Prevent_unnecessary_notifications_from_being_sent_out.js
@@ -1,16 +1,12 @@
-// Create a GlideRecord object for the 'sys_email' table
 var emailGR = new GlideRecord('sys_email');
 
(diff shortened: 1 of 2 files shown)


Lacah:
Thanks for your contribution. Since you are making all the same updates to the GlideRecords, have you thought about using the [updateMultiple()](https://developer.servicenow.com/dev.do#!/reference/api/xanadu/server/no-namespace/c_GlideRecordScopedAPI#r_ScopedGlideRecordUpdateMultiple?navFilter=updateMultiple) method instead :)

Background Scripts/Prevent unnecessary notifications from being sent out/Prevent_unnecessary_notifications_from_being_sent_out.js
```diff
@@ -0,0 +1,16 @@
+// Create a GlideRecord object for the 'sys_email' table
+var emailGR = new GlideRecord('sys_email');
+
+// Query for the emails you want to ignore (adjust the query as needed)
+emailGR.addQuery('state', 'ready'); // only those mails which are ready to send
+emailGR.addEncodedQuery("sys_created_onONToday@javascript:gs.beginningOfToday()@javascript:gs.endOfToday()"); // Optional query to set timeline if not required we can comment this.
+emailGR.query();
+
+// Loop through the results and mark them as 'Ignored'
+while (emailGR.next()) {
```
Lacah, line 10: Sorry I wasn't entirely clear - when you use updateMultiple(), you don't need to loop through the records - that's the point! You send the query **AND** the update to the database in 1 call! So you're code would like like this:
```javascript
// Create a GlideRecord object for the 'sys_email' table
var emailGR = new GlideRecord('sys_email');

// Query for the emails you want to ignore (adjust the query as needed)
emailGR.addQuery('state', 'ready'); // only those mails which are ready to send
emailGR.addEncodedQuery("sys_created_onONToday@javascript:gs.beginningOfToday()@javascript:gs.endOfToday()"); // Optional query to set timeline if not required we can comment this.

//Set any results to ignored
emailGR.state = "ignored"; //setting state to "ignored"
emailGR.type = 'send-ignored'; // Set the type to 'ignored'
emailGR.updateMultiple(); // Save the changes
``` 

Yeah, you don't even need the `query` in this case, due to what I mentioned above!

Lacah (changes requested):
See in-line comment

Lacah (approved):

Lacah:
Thanks!

PR #1527, opened 2024-10-28
 AutoResolveIncident Business Rule

MYaswanth28:
AutoResolveIncident Business Rule
This Business Rule automatically updates the "State" field of an Incident to "Resolved" when the "Resolution Notes" field is filled.


Lacah:
I am not sure this makese sense or adds much value, as it is not how incidents management should work. Also, the script is a one liner (and in fact, does not even need to be scripted), which is one more reason I am not able to accept this.

PR #1528, opened 2024-10-28
Catalog client scripts

MYaswanth28:
# Dynamic Delivery Date Catalog Client Script

This Catalog Client Script dynamically shows or hides the "Delivery Date" field based on the selected value of the "Delivery Method" field. If the "Delivery Method" is "Express", the "Delivery Date" field is shown; otherwise, it is hidden.


Lacah:
I am sorry but again, this is a very basic script that has no real added value and does not align with the purpose of this repository.

PR #1530, opened 2024-10-28
Cmdb ci lifecycle management UI page

Pradeep567iaf:
This UI Page was developed to streamline the update process of the Install Status field for CMDB (Configuration Management Database) items in ServiceNow. Using Jelly XML scripting, the UI Page fetches details for a selected CI item, allows users to modify the Install Status, and save changes back to the CMDB record. Client scripts and script include are used to create a backend logic for fetching configuration item details and client script execute the backend function via GlideAjax api.


Lacah (approved):
Grat stuff

PR #1533, opened 2024-10-28
This  script is designed to facilitate the bulk assignment of a specific role to multiple users at a time.

ShakthiSn:
This  script is designed to facilitate the bulk assignment of a specific role to multiple users at a time.


Lacah:
Thanks for your contribution, though I fail to see why should anybody use this script as is. Roles should be assigned via groups, especially when there are several users involved. I am not sure that this is addressing any gaps in functionality or if it is adding value in any meaningful way.
Therefoe I am closing this PR now.

PR #1535, opened 2024-10-28
Adding placeholder text on Resolution notes Client Script

prash5595:
I added the folder code-snippets->client scripts -> Adding Placeholder on Resolution Notes

I’ve implemented a client script that adds placeholder text in the resolution notes field when the state changes to Resolved. The placeholder guides fulfillers to maintain consistency and disappears as soon as they start typing. This ensures that essential information is not missed while resolving tickets.


snamjosh:
@prash5595 Thanks for your contribution. IS there any particular reason of using g_form.getElement() method instead of g_form.getValue() ?

prash5595:
@snamjosh Thanks for reviewing my contribution. g_form.getValue() is used to retrieves the current value of the field. In my case, I want to add the placeholder text on resolution notes not to get the value of resolution notes. 

Hope it clarifies!

PR #1542, opened 2024-10-28
Fix Script to Remove Leading and Trailing Spaces from a field in bulk records

Adil404:
I have added a fix script to help get rid of removing spaces from leading and trailing manually for a field from thousands of records. 

This fix script will help in removing leading and trailing spaces from a field in bulk records. 
to give an example I have used incident table and short description as a field, you can modify it based on your requirement.


am-shakeel (approved):
Looks good

PR #1546, opened 2024-10-28
Open Record producer from Workspace and auto-populate data on the opened record producer

rohi-v:
When we want to open a catalog item or record producer with details from current record and map it to the opened catalog form then we can use this code.

This UI Action and catalog client script Redirects you to the record producer or catalog item( based on the sys id provided) and auto-populates the fields from the parent record to the catalog item/record producer variables.


am-shakeel (approved):
Looks good

PR #1550, opened 2024-10-29
Fetching a reference field value from higher-level parents when there is a field containing the parent record in the children and the use-case reference field is present in all tables 

Soumyadeep10:
In this before insert or update Business Rule, we are fetching a reference field value from higher-level parents in hierarchy when there is a field containing the parent record in the children and our use-case reference field is present in all the tables in hierarchy


kmohit2809:
@Soumyadeep10 thanks for your contribution. However, there are few suggestions. Please correct the name of readme.md file (it contains small 'm'). Make sure that readme.md describes your snippet clearly. PR name can be more brief. Please correct these things and and recommit them.

PR #1556, opened 2024-10-29
Decryption in ServiceNow Using AES 256

TheShashank13:
This code can be used to generate JAR file which can be used for decription of payload using AES 256 key Decryption.Java file has to be convereted to JAR and needs to called via ECC queue and MID Server ServerScript Method.Js method contains the details of the code to trigger server script A system property named 'decrypt_key' needs to created to store AES decrytion key

### DIFF
--- a/batman.html
+++ b/batman.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en" >
+<head>
+  <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


am-shakeel (changes requested):
There are NO files in this PR

TheShashank13:
yes, it is deleted I have submitted #1559 as replacement for the same

PR #1557, opened 2024-10-29
Set update sets to complete

koratalas:
Update sets which are inprogress state shall be set to complete using this background script if it matches the below conditions:
1) State should be inprogress
2) Name should not contain "default"

This helps the administrators to set the update sets state to complete state.


SapphicFire (approved):
Looks good to me. Consider adding more functionality and guidance about ways to apply your snippet in future submissions

PR #1558, opened 2024-10-29
Pass multiple values from server to client side in an object

NowRishi:
Created a client callable script include and onChange client script to get multiple parameters from server to client in an object and then setting those values in appropriate fields using the onChange client script.


SapphicFire (approved):
Looks good to me

PR #1560, opened 2024-10-29
Generated SQL query checker

manrick:
There is a simple way, how the generated SQL query can be seen on the output without enabling the SQL Debug feature in the navigator. This pull request contains the solution of this feature.


SapphicFire (approved):
Looks good to me. It may be worthwhile calling out the risks of leaving trace on as well as the nature of that method in more detail in future

PR #1562, opened 2024-10-29
Encrypt and Decrypt Password Fields

vamsi0777:
Dear ServiceNow Community,

The GlideEncrypter API uses 3DES encryption standard with NIST 800-131 A Rev2 has recommended against using to encrypt data after 2023. ServiceNow offers alternative cryptographic solutions to the GlideEncrypter API. 

Glide Element API to encrypt/decrypt password2 values through GlideRecord.

Below are the sample scripts I ran in my PDI: For Password fields. 

Note: 'u_pass' is Password (2 Way Encrypted) field.


SapphicFire (approved):
Looks good to me

PR #1569, opened 2024-10-30
Valid mobile number

salma-93:
It is used to validate the length of the phone number. If it exceeds the maximum length, it displays an error message and clears the field.

### DIFF
--- a/Client Scripts/validate phone number/validateLength.js
+++ b/Client Scripts/validate phone number/validateLength.js
@@ -1,0 +1,13 @@
+// Client Script: Validate Length
+// Table: sys_user
+// Type: onChange
+// Field: phone
+
+function onChange(control, oldValue, newValue, isLoading) {
+    if (isLoading || newValue === '') return;
+
+  if (newValue.length>12) {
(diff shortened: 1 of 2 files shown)


Client Scripts/validate phone number/validateLength.js
```diff
@@ -0,0 +1,13 @@
+// Client Script: Validate Length
+// Table: sys_user
+// Type: onChange
+// Field: phone
+
+function onChange(control, oldValue, newValue, isLoading) {
+    if (isLoading || newValue === '') return;
+
+  if (newValue.length>12) {
```
wiz0floyd, line 9: What if somebody uses spaces or dashes in the phone number? `+1 (800) 123-4567` is 17 characters and is a valid format for a phone number in the US. Have you considered using a regex to test the phone number?  Additionally there is a dedicated field type for phone numbers, so can you provide some context around when you would use this script instead of the field?

wiz0floyd (changes requested):

wiz0floyd:
I've requested changes in the script. I'm concerned that this is too simple to include in the code snippets repo as length is not a very robust method for validation of a phone number.

PR #1571, opened 2024-10-30
Fetching reference field value from higher-order parents

Soumyadeep10:
This is a "before insert/update" Business Rule We are fetching a reference field value from higher-level parents in hierarchy when there is a field containing the parent record in the children and our use-case reference field is present in all the tables in hierarchy

In the code, we are referring to "reference field name we want to populate" as "r1" In the code, we are referring to "reference field containing parent record" as "parent"

The "JSUtil.nil" is being used to check for empty/null value for the field.

Through the code we are checking the empty value of the use-case reference field and dot walking to parents and fetching the value from them if it exists


wiz0floyd:
Thank you for your contribution. However, it doesn’t comply with our contributing guidelines. Please move your snippet into its own subfolder. As a reminder, the general requirements (as outlined in the [CONTRIBUTING.md file](https://github.com/ServiceNowDevProgram/code-snippets/blob/main/CONTRIBUTING.md)) are the following: follow the folder+subfolder guidelines and include a readme.md file explaining what the code snippet does. Review your contribution against the guidelines and make the necessary adjustments. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

PR #1573, opened 2024-10-30
UX Data Broker Transform Starter Template

yansa-reece:
This pull request contains a starter template for UX Data Broker Transform scripts. The template code includes:

JSDoc type annotations
ES6 features: destructured inputs, const/let, interpolated strings
Error logging and propagation through try/catch
An example properties array
A readme.md file is also provided that details the template


SapphicFire (approved):

PR #1574, opened 2024-10-30
UX Data Broker Transform Create/Update User Preference

yansa-reece:
This pull request contains code for a transform data broker that creates or updates a user preference for the current user. The data broker accepts the name of the preference and the value to set it to as inputs.


SapphicFire (approved):

PR #1575, opened 2024-10-30
UX Client Script Debug Event

yansa-reece:
This contribution provides a simple UX Client Script called Debug Event, designed to log event details to the console for easy debugging within UI Builder. By displaying event properties like elementId and name, this script helps developers track interactions and troubleshoot event-driven functionality in real time. It may seem simple but this is a utility that I add to every UI Builder project.


SapphicFire (approved):

PR #1576, opened 2024-10-30
UX Client Script Debug State

yansa-reece:
This contribution provides a Debug State UX Client Script for logging the current client state to the console in ServiceNow. Designed to simplify debugging, this script gives developers easy access to the full state object, enabling quick inspection of state values and more effective troubleshooting within UI Builder. It may be simple but its a helpful utility that I add to every UI Builder project.


SapphicFire (approved):

PR #1582, opened 2024-10-30
Set Field 

TechnologistTim:
This will populate a <u_location> field string based on the <location> field on the user's record without having to use GlideAjax. Go to the <u_location> dictionary and place this snippet in the Default Value field.

### DIFF
--- a/Flow Actions/Set Location/Readme.ms
+++ b/Flow Actions/Set Location/Readme.ms
@@ -1,0 +1,1 @@
+This will populate a <u_location> field string based on the <location> field on the user's record without having to use GlideAjax. Go to the <u_location> dictionary and place this snippet in the Default Value field.
--- a/Flow Actions/Set Location/Set Field
+++ b/Flow Actions/Set Location/Set Field
@@ -1,0 +1,1 @@
+javascript: var luser = new GlideRecord("sys_user"); luser.get("sys_id", gs.getUserID()); luser.location.u_installation_id.sys_id.toString();


sandeepd26:
Hi @TechnologistTim ,
Are you asking to put this condition as a flow action?

Flow Actions/Set Location/Set Field
```diff
@@ -0,0 +1 @@
+javascript: var luser = new GlideRecord("sys_user"); luser.get("sys_id", gs.getUserID()); luser.location.u_installation_id.sys_id.toString();
```
SapphicFire, line 1: This call type would not be suitable in a flow action. Additionally, you should use `getUniqueValue()` to get the sys_id string

Flow Actions/Set Location/Readme.ms
```diff
@@ -0,0 +1 @@
+This will populate a <u_location> field string based on the <location> field on the user's record without having to use GlideAjax. Go to the <u_location> dictionary and place this snippet in the Default Value field.
```
SapphicFire, line 1: Invalid file type, `u_installation_id` from the script is not documented, and this readme does not describe much at all

SapphicFire (changes requested):

Flow Actions/Set Location/Readme.ms
TechnologistTim: Sorry, I looked at another readme from someone else and though I followed what they did. I'll add more detail. 

As for the filetype, I followed the contribution file but for a newb like me, it doesn't provide the level of detail I need. I simply added a file as it states and typed away.

As for the field. I'm going to add in the Readme this applies to a reference field that is added. "u_installation" being the name of my particular field. The <> was an attempt to say "whatever you want to name it".

Flow Actions/Set Location/Set Field
TechnologistTim: It works flawlessly on our production instance as is. I did mean to add this to Dynamic Filter though, not Flow Action.

PR #1609, opened 2024-10-31
Reject Approvals Older Than One Year

pallipoina92:
The overall purpose of this script is to identify and reject approval requests that are older than one year and are still in the "requested" state. This helps in cleaning up old approvals that may no longer be relevant.

### DIFF
--- a/Scheduled Jobs/Reject approvals created before an year/readme.md
+++ b/Scheduled Jobs/Reject approvals created before an year/readme.md
@@ -1,15 +1,7 @@
-**Creating a GlideRecord Object:**
-var grAppr = new GlideRecord("sysapproval_approver");
+**Script Purpose:**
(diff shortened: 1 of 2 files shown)


Scheduled Jobs/Reject approvals created before an year/readme.md
Lacah, line 1: This is not a real readme, only the same code with some comments (which you could just add to the actual js file).
Could you add some actual description here about the purpose of the script and what should someone who wants to use it do?
Thanks.

Lacah (changes requested):

pallipoina92:
Hi,

Sure, will do that.

Thanks,
Sunil

Scheduled Jobs/Reject approvals created before an year/readme.md
pallipoina92: Hi, I have updated the readme.md file and committed the changes, is there anything else that I need to do?

Lacah: This is sufficient, thanks.

Lacah (approved):

PR #1614, opened 2024-11-01
Client scripts

TechnologistTim:
Changing a boolean field true or false based on a role the signed in user has. I personally used it if a clerk had a type of clerk role to change it to true so I can use that as a condition for further features like setting ACLs based on that field being true.


Client Scripts/Set field based on roles/content
SapphicFire, line 1: Empty file

Client Scripts/Set field based on roles/Set field based on role.js
SapphicFire, line 1: Near duplicate of [this](https://github.com/ServiceNowDevProgram/code-snippets/blob/25796aeb153cd6093ca96a894109567551df6e89/Client%20Scripts/Enable%20VIP%20and%20Senior%20VIP%20Checkboxes%20and%20read%20only/ScriptToEnableVIP_superVIP%20users#L4) snippet

SapphicFire (changes requested):
Thanks for your submission. It appears that this pull request contains more files than described in the title. Please review and update the pull request to either remove the extra/unintended files or to update the description to be more comprehensive. Ensure your contributions are meaningful and adhere to the project’s standards. Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

This is also a duplicate

PR #1615, opened 2024-11-01
Trigger Event when someone is added to a list

TechnologistTim:


Business Rules/Trigger Event when a member is added to a list/Trigger an event when user is added to field.js
```diff
@@ -0,0 +1,36 @@
...
+        // Loop through the added users to trigger the event for each
+        addedUsers.forEach(function(userID) {
+            var email;
+            var firstName;
+
+            // Try to get the user record by user ID (sys_id)
+            var userGr = new GlideRecord('sys_user');
+            if (userGr.get(userID)) {
+                firstName = userGr.first_name;
+                email = userGr.email;
```
SapphicFire, line 24: In future, look to use getters such as `getValue('field')` instead of getting the GlideElement each time

SapphicFire (approved):
Looks good to me, incorporated some feedback for future consideration

Business Rules/Trigger Event when a member is added to a list/Trigger an event when user is added to field.js
TechnologistTim: Thank you for the tip.

PR #1616, opened 2024-11-01
Mail script

TechnologistTim:
A mail script that includes a parsed first name, a subject, case details, and attachments.


Mail Scripts/Added a case/content
SapphicFire, line 1: Invalid file

Mail Scripts/Case details/Case details notification.js
```diff
@@ -0,0 +1,110 @@
+(function runMailScript(current, email, email_action, event) {
+    var gr = new GlideRecord('sn_customerservice_case'); // Replace with case table name)
```
SapphicFire, line 2: You already have current, why are you getting it again

Mail Scripts/Case details/Case details notification.js
```diff
@@ -0,0 +1,110 @@
+(function runMailScript(current, email, email_action, event) {
+    var gr = new GlideRecord('sn_customerservice_case'); // Replace with case table name)
+    var caseLink = '<a href="' + current.getLink() + '" style="font-weight: bold;">' + current.number + '</a>'; // Creates a clickable link to the case record
+    var subject = 'Added to case ' + current.number; // Sets the subject with the current case number
```
SapphicFire, line 4: Use getters (`getValue('field')`) instead of GlideElements and relying on string casting

Mail Scripts/Case details/Case details notification.js
```diff
@@ -0,0 +1,110 @@
+(function runMailScript(current, email, email_action, event) {
+    var gr = new GlideRecord('sn_customerservice_case'); // Replace with case table name)
+    var caseLink = '<a href="' + current.getLink() + '" style="font-weight: bold;">' + current.number + '</a>'; // Creates a clickable link to the case record
+    var subject = 'Added to case ' + current.number; // Sets the subject with the current case number
+    email.setSubject(subject);
+
+    // Querying for the case record
+    gr.addQuery('sys_id', current.sys_id);
```
SapphicFire, line 8: `getUniqueValue()` does this but again you don't need this additional GlideRecord at all

SapphicFire (changes requested):
Thanks for your submission. It appears that this pull request contains more files than described in the title. Please review and update the pull request to either remove the extra/unintended files or to update the description to be more comprehensive. Ensure your contributions are meaningful and adhere to the project’s standards.

I will leave this open for a short period if you wish to change this.

Mail Scripts/Added a case/content
TechnologistTim: Deleted "contents". Is it necessary to make a placeholder file to create the folder then go back and delete it?

Mail Scripts/Case details/Case details notification.js
TechnologistTim: I thought it was necessary when calling a field. So, I can use getLink(), number, sys_id etc without using current.?

Mail Scripts/Case details/Case details notification.js
TechnologistTim: Removed.

Mail Scripts/Case details/Case details notification.js
TechnologistTim: Just simplified it.

Mail Scripts/Case details/Case details notification.js
TechnologistTim: Done.

Mail Scripts/Added a case/content
SapphicFire: Not strictly. While you do need a file in a folder for GitHub to accept it, you can make this be one of the files you plan to keep like readme.md

Mail Scripts/Case details/Case details notification.js
SapphicFire: For future reference, that GlideRecord isn't necessary because that's what is provided through the current object. In doing this you add another call to the database only to get the same thing you already had.

Additionally, the variable name `gr` carries risks because it's used a lot in ServiceNow and unless you wrap your code in a function, it can collide with other declared instances of gr. While you are not at that specific risk here, that's a good practice to change

Mail Scripts/Added a case/content
SapphicFire, line 1: Is this a new file? Unfortunately this one needs to go too

SapphicFire (changes requested):

Mail Scripts/Added a case/content
TechnologistTim: Once I add "mail script/" I don't see Add File. That's why I was going that route.

PR #1617, opened 2024-11-01
Created a script to get the count of incidents by priority under each category.

vijaytalupula:
Created below files as part this pull request,

1. readme.MD
2. code.js

### DIFF
--- a/GlideAggregate/List the incident priority count under each category/code.js
+++ b/GlideAggregate/List the incident priority count under each category/code.js
@@ -1,0 +1,20 @@
+// Get the count of Incidents by Category and then by Priority.
+
+var incCATGR = new GlideAggregate('incident');
+incCATGR.addAggregate('COUNT', 'category');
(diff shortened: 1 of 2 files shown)


GlideAggregate/List the incident priority count under each category/code.js
```diff
@@ -0,0 +1,20 @@
+// Get the count of Incidents by Category and then by Priority.
+
+var incCATGR = new GlideAggregate('incident');
+incCATGR.addAggregate('COUNT', 'category');
+incCATGR.orderBy('category');
+incCATGR.query();
+
+while (incCATGR.next()) {
+	var cat = incCATGR.category;
```
SapphicFire, line 9: You don't use this variable, and additionally should use getters to avoid relying on GlideElement casting

SapphicFire (approved):
Looks good to me, I have included some feedback to consider in the future

PR #1646, opened 2025-10-01
Auto Disable the Service Accounts due to inactivity

at8807602:
This schedule job script will help to auto disable the service accounts which has not been logged in or used for last 30 days. This can be helpful for those who are looking to disable the accounts based on some certain time period.

### DIFF
--- a/Server-Side Components/Scheduled Jobs/Auto Disable account/Disable Service accounts due to inactivity
+++ b/Server-Side Components/Scheduled Jobs/Auto Disable account/Disable Service accounts due to inactivity
@@ -1,6 +1,6 @@
 var user = new GlideRecord("sys_user");
 user.addActiveQuery();
-user.addEncodedQuery("last_login!=NULL");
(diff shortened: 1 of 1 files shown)


ravichandra1998g (changes requested):
Hello Abhishek,

Transaction logs table is one of the largest tables in the instance which stores all the user/background transactions. 

querying it for the last 40 days would have performance implications on the system. 

Can you optimize the script for the performance?

at8807602:
Hello @ravichandra1998g ,

Thanks for the validation. As requested, I have made some certain changes by optimizing the performance impact. Not this looks good from my end in terms of system performance. Please review it and let me know, if any changes are required.

ravichandra1998g:
This applies to all the users as there is no filter to include only service accounts.. please add the filter.

at8807602:
Hello @ravichandra1998g ,

Again, thanks for the validation. I have applied the filter for only service account. Please validate it and let me know, if anything needs to be changed.

PR #1673, opened 2025-10-02
CMDB/IRE - CI deduplication task generator

JonSandoval:
The script rechecks the cmdb_ci_hardware table (or any configured CI table) for records that share the same serial number. This is for identifying duplicates that may not have been processed by IRE or imports that did not run through IRE.


wiz0floyd (approved):

wiz0floyd:
Looks good!

PR #1733, opened 2025-10-03
Create CloneUser.js

Thrizvi:
This script is designed to duplicate an existing user record in ServiceNow using the GlideRecord API. It is useful for scenarios where you need to replicate user profiles for testing, onboarding templates, or automation purposes.


Thrizvi:
is this correct now

PR #1785, opened 2025-10-04
User acknowledgement using u iscript and userpreferences

gowdah:
**# PR Description:** 

This is to display an acknowledgement message when a user logs into the system, using a UI script, UI page, and a user preference.

The UI script checks the user preference and calls the UI page upon user login.
The UI page displays the message content.
The user preference is used to acknowledge the message and prevent it from showing every time the user logs in.


ravichandra1998g (approved):
Thank you for submitting! looks good!

PR #1805, opened 2025-10-05
Track tag removal using delete business rule

gowdah:
**# PR Description:** 

This is a delete business rule on the "**label_entry**" table that tracks tag removal or deletion and updates audit entries on the record."


ravichandra1998g (approved):
Thank you for submitting!
looks good!

PR #1808, opened 2025-10-05
Group similar assessments using flow designer action

gowdah:
**# PR Description:** 

This is a Flow Designer action that takes similar assessment **sys_ids** as input, verifies them and groups the assessments if the grouping criteria are met.


ivan-betev:
Thank you for you contribution, however, your code is not following the best practice, as I can see issues in the inputs declaration area, nested GlideRecord query in a while loop, inappropriate usage of initialize method. That might lead to an unpredictable behavior. Please fix the code and resubmit or reopen. Closing this for now.

PR #1811, opened 2025-10-05
Update Field Based on Another Table Data

chetnadev:
My first 2025 Hacktoberfest submission
The basic use case of this script is Let’s say you want each incident record to store the caller’s department, but this info is only on the user profile.
This script pulls it from the user and updates all related incident records.

### DIFF
--- a/Server-Side Components/Background Scripts/Update Field Based on Another Table Data/README.md
+++ b/Server-Side Components/Background Scripts/Update Field Based on Another Table Data/README.md
@@ -1,0 +1,9 @@
+This script Update incidents with the department of their caller
+Loops through every user in the sys_user table.
+For each user, finds all incidents where they are the caller (caller_id matches sys_id of the user).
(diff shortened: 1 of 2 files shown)


rohi-v:
Hi @chetnadev 

Instead of querying the user table (which can be quite large depending on the number of users) to find incidents related to a user, we can directly query the Incident table and dot-walk to the user’s department for mapping.
Also, since Department is already dot-walkable from the user record, it can be displayed directly on the form without creating a custom field.

Just a quick note: this repository already contains similar examples, so please check for duplicates or overlapping use cases before submitting.

Closing this for now. Once you make the suggested changes, feel free to re-open this Pull Request or create a new one.

rohi-v:
Hi @chetnadev 

we appreciate the effort you’ve put into this contribution. Since this is your first PR, we’ve labeled it as hacktoberfest-accepted  so it still counts toward your Hacktoberfest progress. Thank you for contributing!

PR #1823, opened 2025-10-05
Use system property as an object to store multiple values and retrieve attributes when needed

gowdah:
# PR Description: 

This is to use a **string** system property and convert it to an **object** and retrieve object attributes when required.

**Example use case:** This comes in very handy when we need to retrieve static IDs for different instances.


WillemZeiler (approved):

WillemZeiler:
Looks good!

PR #1842, opened 2025-10-06
Updated Ulaction.js: Added modal function to edit last Work Notes or Comments

Shweyy123:
📝 Pull Request Description

This update to UIAction.js introduces a new client-side function openEditLastCommentModal() that enables users to open a modal for editing the most recent Work Notes or Additional Comments on the Incident form.

🔄 Key Changes:

1) Added a documented function openEditLastCommentModal().

2) Utilizes GlideModal to launch a UI macro (edit_worknotes_comments_inc) in a dialog.

3) Passes the current record's sys_id using setPreference("incid", g_form.getUniqueValue()).

4) Sets the modal width to 550 and applies a custom title.

5) Designed for use with a UI Action configured as:

Client: true

Form Button: true

OnClick: openEditLastCommentModal()

📈 Purpose:

To enhance the user experience by allowing inline editing of the last communication fields without leaving the Incident form.

### DIFF
--- a/Client-Side Components/UI Pages/Edit Last WorkNotes/UIaction.js
+++ b/Client-Side Components/UI Pages/Edit Last WorkNotes/UIaction.js
@@ -1,10 +1,28 @@
-//Enable client set to true
-//Enter Onclick value as function name openEditLastCommentModal()
-//Enter form button as true
+/**
+ * Opens a modal to edit the last Work Notes or Additional Comments.
+ * 
+ * @description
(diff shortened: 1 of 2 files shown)


shreyawani98:
Hi @Shweyy123 

After reviewing your request, it seems that you have not added the ReadMe.md file. Apart from that, since the use case is about the WorkNotes, please exclude the other javascript file about HrTask and keep only the WorkNotes UIaction javascript file.

Please submit another pull request as mentioned above.

Thanks

PR #1879, opened 2025-10-06
Fix Corrupted Catalog Item References in RITMs

chetnadev:
The purpose of this script in ServiceNow is to correct data inconsistencies in the Requested Items (sc_req_item) table by 
identifying and updating records that are missing a reference to their associated catalog item (cat_item). Specifically, it searches for RITM records where the cat_item field is null, which typically indicates an incomplete or improperly imported record. For each of these, the script attempts to find a matching catalog item in the sc_cat_item table by comparing the RITM's short_description to the catalog item's name.
If a match is found, the script updates the RITM to reference the correct catalog item and keeps a count of how many records were fixed. This automated cleanup helps restore proper data relationships between requested items and their catalog definitions, which is important for reporting, workflows, and overall data integrity in the ServiceNow instance.


snamjosh:
Thank you for your contribution. But, please find better solution to address this issue since it's not a good idea to lookup records using short description of catalog item. OOB, ITIL users can modify description of sc_req_item records which will break this proposed fix.

PR #1887, opened 2025-10-06
Client Script - Incident to Problem Fields Mapping

mitalizope:
Client Script - Maps Incident Fields to Child Problem
Description:
This client script automatically maps key field values from a parent Incident record to a Problem record when the user creates a new Problem and selects an Incident record in the Parent reference field.

Client Script:
Name: IncidentToProblemMapping
Table: Problem [problem]
Type: onChange
Field name: Parent

Script Include:
Name: IncidentDetails
Glide AJAX enabled: checked

Implementation:
Place this script on the Problem table with the condition that it triggers on change of the Parent field. When the selected parent record is from the Incident table, it fetches field values from the Incident using GlideAjax and sets them on the Problem form.
Fields Mapped:
•	Configuration Item
•	Priority
•	Assignment Group
•	Short Description
•	Description

### DIFF
--- a/Client-Side Components/Client Scripts/Incident to Problem Field Mapper/IncidentDetails.js
+++ b/Client-Side Components/Client Scripts/Incident to Problem Field Mapper/IncidentDetails.js
@@ -1,0 +1,22 @@
+//Script include to get the values of incident record fields
+var IncidentDetails = Class.create();
+IncidentDetails.prototype = Object.extendsObject(AbstractAjaxProcessor, {
+
+	getIncidentDetails: function() {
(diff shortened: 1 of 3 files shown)


WillemZeiler:
Thank you for your submission. I will mark it to get you points for Hacktoberfest. However, please note that OOTB ServiceNow allows for creating a problem from an incident. We advice to use that instead of custom scripts. Example for workspace: https://www.servicenow.com/docs/bundle/washingtondc-it-service-management/page/product/itsm-workspace/task/create-problem-from-incident.html.

The PR will be closed without merging. But your contribution counts towards the hacktoberfest.

PR #1924, opened 2025-10-07
Return date validation for a catalog item using onChange script

Soumyadeep10:
This piece of code runs in an OnChange catalog client script for the field 'u_return_date' and validates the return date(u_return_date) in a ServiceNow Catalog item form to ensure:

1. A start date(u_start_date) is entered before setting a return date(u_return_date).
2. The return date is within 6 months after the start date.
3. Return date is not the start date or before that.

Let’s say with an example:

  u_start_date = 2025-10-01
  You enter u_return_date = 2026-04-15
  
  Steps:
  a)Difference = 196 days → More than 180 days
  b)Result: u_return_date is cleared and error shown: “Select Return Date within 6 months from Start Date”


Soumyadeep10:
Hello @WillemZeiler , I have made the changes requested by you.

PR #1950, opened 2025-10-08
Automatically Reopen Incidents Closed Prematurely

chetnadev:
If a user reopens a closed incident within 5 minutes of closure (e.g., because the issue wasn't actually resolved), automatically reopen it, log the reason, and notify the assigned user.
This code detects when an incident is reopened within 5 minutes of being closed. It compares the previous and current state of the record, and if it was previously Closed (state = 7) and is now Active, it calculates the time since closure. If the reopening happened within 5 minutes, it:
Triggers an event incident.reopened_quickly for notifications or logging.
Adds a work note explaining the automatic flag.
Sets a custom flag field u_reopened_flag to true for tracking.
This is useful for identifying and tracking incidents that are quickly reopened, possibly indicating incomplete resolution.


WillemZeiler:
Thank you for your submission. I will mark it to count towards Hacktoberfest. However, we will not merge it, as the described functionality is available OOTB. On a notification configuration you can send out an email on reopening. State and time per state can also be tracked and reported on using Metrics.
We recognize the efforts, so you do get points!

PR #1964, opened 2025-10-09
Service Request Submission Throttling

chetnadev:
This business rule in ServiceNow is designed to limit the number of service catalog requests a user can submit within a certain timeframe specifically, 5 requests per hour.

This rule helps throttle excessive catalog request submissions, which can:
Prevent system abuse (intentional or accidental),
Avoid spamming by users (especially in large environments)

This ServiceNow Business Rule checks if the current user has submitted 5 or more service catalog requests (sc_request) in the past hour. If they have:
It shows an error message, And aborts the current request from being submitted.
It uses GlideAggregate to efficiently count recent requests made by the same user, and compares the count against the limit.


rohi-v:
Hi @chetnadev 

Please mention the BR details(table Name, When to Trigger) in the readme file.

The logic doesn't seem to work when validated from portal and native UI too.

Closing this for now. Once you make additional changes, feel free to re-open this Pull Request or create a new one.

PR #2037, opened 2025-10-11
Switching the unassignment automation to use the central script from Hacktoberfest repo

Lacah:
Earlier I committed the script to the Hacktoberfest repo, in an attempt to just keep one version of it centrally. Which this PR, I am replacing the previous automation in code-snippets with a shim that simply references the central, and hoping that it works. If it does, I plean to do the same for other repos.


WillemZeiler:
Good idea, looks cleaner.

PR #2051, opened 2025-10-12
Send Custom Slack Message When VIP Logs an Incident

chetnadev:
When a VIP user creates an incident, notify a Slack channel (or send via webhook) with a custom JSON message.
Note : You’d need to set up a Slack Incoming Webhook for this.

This business rule triggers on incident creation and checks if the `caller` has the `vips` flag set to `true`. If so, it constructs a JSON payload with incident details and sends a POST request to a predefined Slack webhook URL using `sn_ws.RESTMessageV2`. The purpose is to notify support teams in real time when a VIP user logs an incident.

### DIFF
--- a/Server-Side Components/Business Rules/Send Custom Slack Message When VIP Logs an Incident/README.md
+++ b/Server-Side Components/Business Rules/Send Custom Slack Message When VIP Logs an Incident/README.md
@@ -1,0 +1,4 @@
+When a VIP user creates an incident, notify a Slack channel (or send via webhook) with a custom JSON message.
+Note : You’d need to set up a Slack Incoming Webhook for this.
+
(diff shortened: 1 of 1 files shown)


rohi-v:
Hi @chetnadev 

Thank You for your contribution

The REST message call using sn_ws.RESTMessageV2 is a standard approach already available within the platform, so this doesn’t add much as a unique use case.

Additionally, there’s no error handling implemented for the REST call (for example, handling failed responses or exceptions).

Code snippets reused from the [ServiceNow Documentation](https://docs.servicenow.com/) or [API References](https://developer.servicenow.com/dev.do#!/reference/) are acceptable only if they are expanded in a meaningful way (e.g., with additional context, documentation, or variations). Remember: "QUANTITY IS FUN, QUALITY IS KEY."

PR #2053, opened 2025-10-12
Auto Create Problem Records for Recurring Incidents

chetnadev:
This "after" business rule automatically creates a Problem record when a particular Configuration Item (CI) has had 5 or more incidents in the last 24 hours, and no open Problem already exists for that CI.
This helps in proactive problem management, aiming to address recurring issues.
Here’s the working of the code explained:

 - Check if CI is present in the current Incident (current.cmdb_ci).
 - Count incidents created in the last 24 hours for the same CI using GlideAggregate.

If 5 or more incidents are found for that CI:
 - Query the Problem table to check if an open Problem (not closed) already exists for that CI.
 - If no open Problem exists, create a new Problem record with: The same CI, A predefined short description And set its state to New (1).
 - Log a message indicating that a Problem has been created.
This automates Problem creation for frequent incidents on the same CI.

### DIFF
--- a/Server-Side Components/Business Rules/Auto Create Problem Records for Recurring Incidents/README.md
+++ b/Server-Side Components/Business Rules/Auto Create Problem Records for Recurring Incidents/README.md
@@ -1,6 +1,12 @@
-This business rule automatically creates a Problem record when there are 5 or more incidents linked to the same Configuration Item (CI) within the last 24 hours.
-How it works technically:
--It first checks if the current Incident has an associated CIThen, it counts the number of incidents created in the last day for that CI.
--If the count is 5 or more, it looks for any open Problem records related to the same CI.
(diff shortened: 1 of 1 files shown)


ivan-betev:
Thanks for your contribution, you are almost there.

Please invest some time into a proper README formatting otherwise it's not user-friendly, doesn't explain what type of business rule and the conditions user should user.

In your code you are using gr as a variable name, which is dangerous in a global scope.

Please fix this and let me know.

chetnadev:
Hi @ivan-betev . I've done the suggested changes in both files, please check and let me know.

ivan-betev:
Thanks for your contribution @chetnadev, I'll approve it, however, please avoid using unclear variable names (ck ???) and try to follow the best practice.

PR #2277, opened 2025-10-18
Dynamic Field Helper — Auto Tooltip Enhancer for Forms

abhix2112:
##Dynamic Field Helper — Auto Tooltip Enhancer for Forms

### Overview:
The Dynamic Field Helper automatically adds contextual tooltips beside field labels, explaining field purpose, business rules, or dependencies.

Features:

No form edits required — works dynamically

Pulls data from GlideAjax or local config

Lightweight and visually integrated with the platform

Great for onboarding or complex form guidance

Use Case:
Improve user understanding and reduce data-entry errors by providing inline guidance without cluttering forms.

### DIFF
--- a/Specialized Areas/Fix scripts/updateMultipleRecords/README.md
+++ b/Specialized Areas/Fix scripts/updateMultipleRecords/README.md
@@ -1,8 +1,0 @@
-*************
-This script will query the group table and look for groups with inactive members. The script will replace the inactive manager with the oldest active member of the group.
-
-Logs: If no active members are there in group //gs.info("Group " + grp.name + " does not have any active user");
-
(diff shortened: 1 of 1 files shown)


kmxo:
@abhix2112 Thank you for your participation. 

Unfortunately it’s usually not a good idea to use direct DOM manipulation in ServiceNow.

It can sometimes be done carefully for small UI tweaks, but it’s not recommended and often leads to upgrade issues, broken functionality, or performance problems. 

Providing tooltip to help the user is not a bad idea at all, but it is recommended to avoid the 'document' reserved word which is DOM manipulation. 

Please try to use supported APIs like:

`g_form.setHelpText('short_description', 'Describe the problem in a clear way.');`

PR #2398, opened 2025-10-22
TechTrekwithAJ_ITOM Discovery automate Via Workflow script

TechTrekwithAJ:
This script is used in ServiceNow to automate device discovery on a network by triggering Quick Discovery using a MID Server. It is typically used within a workflow or flow, where you want to scan a device using its IP address.
It adds reliability by using two MID servers, so if one fails, the other is used as a backup.

Two MID servers are defined to run the discovery process, providing a fallback if the first one fails.
Replace 'mid_server_1' and 'mid_server_2' with actual MID server names or sys_ids.
The script fetches the IP address either from the current record (current.ip_address) or from the inputs.
If no IP address is available, it logs an error and exits.
A helper function is used to initiate Quick Discovery for a specific IP using a specified MID server.
It first attempts discovery using the first MID server.
If successful, it logs the success and updates the record with the status.
If it fails, it logs a warning and tries the second MID server.
If both attempts fail, it logs an error and marks the discovery as failed.
Finally, it updates the record with the outcome and the MID server used.


ravichandra1998g:
Hello @TechTrekwithAJ 

This cannot be merged as this creates a new dummy path ITOM under code snippets.
we already have ITOM folder under Specialized Areas.

Please DO Not CREATE Duplicate folders.

also, I have just noticed it's calling new Discovery (), but there is no  script include and the method it's calling.
Is there a script include that triggers the discovery?

PR #2504, opened 2025-10-26
check duplicates from server class

d-sharmagit:
# PR Description: 

Scan all Servers (cmdb_ci_server). For each one, check if there is another CI in cmdb_ci_computer with the same name but not a server (sys_class_name != cmdb_ci_server).
If found, log the server name and the duplicate CI’s class; keep a running duplicate count; finally log the total.


bird-03:
Hey @d-sharmagit 
Thanks for your contribution!!

Code looks good and it's useful script in identifying the duplicate servers. We are accepting this for now though recommend to update the description with proper tile as what this PR is about instaed of 'Diksh code hac 06'

Happy Hacking

d-sharmagit:
Udated

bird-03:
Thanks for change!

